PR #37, opened 2014-02-10
Add support for empty string as select key

eemeli:
This makes the following work: `The key is {KEY, select, ''{missing!} other{this: {KEY}}}`

It adds `''` (two single quotes) as a special `stringKey` that matches an empty string. Included are two new tests that show how it works.

The particular expression that inspired this for me was the following, which required separately generating `GOT_*` booleans for each case:

```
Listing { N, plural, one { one item } other { # items } }
{ GOT_DAY, select, true { on {DAY} } other {} }
{ GOT_AREA, select, true { in {AREA} } other {} }
{ GOT_Q, select, true { matching the query {Q} } other {} }
```

which I could now rewrite as follows:

```
Listing { N, plural, one { one item} other { # items } }
{ DAY, select, ''{} other { on {DAY} } }
{ AREA, select, ''{} other { in {AREA} } }
{ Q, select, ''{} other { matching the query {Q} } }
```

### DIFF
--- a/lib/message_parser.js
+++ b/lib/message_parser.js
@@ -778,6 +778,26 @@ module.exports = (function(){
           if (result0 === null) {
             pos = pos0;
           }
+          if (result0 === null) {
+            pos0 = pos;
+            if (input.substr(pos, 2) === "''") {
+              result0 = "''";
+              pos += 2;
+            } else {
+              result0 = null;
+              if (reportFailures === 0) {
+                matchFailed("\"''\"");
+              }
+            }
+            if (result0 !== null) {
+              result0 = (function(offset) {
+                return "";
+              })(pos0);
+            }
+            if (result0 === null) {
+              pos = pos0;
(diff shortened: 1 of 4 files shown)


samuelneff:
#60 doesn't really address the same thing. #60 demonstrates how to match `undefined`. An empty string is not undefined and it would be useful to be able to match an empty string.

kavi87:
It is a bug in my opinion. Our users can write their own keys like `{ firstName } { lastName } - { secretAttribute }` but we do not want "John Doe - undefined" to be printed in some case. And handling the undefined case would be cumbersome in every keys. 

For now my solution is to use the parser and check that the args are in the context and replace absent variable with an empty string. 

A more useful syntax would be something like `{ firstName } { lastName } [[ - { secretAttribute } ]]` which would remove everything between the [[ ]] if one of the args inside is missing.

PR #85, opened 2015-01-13
Parser improvements

vogelsgesang:
I am currently getting started with MessageFormat and as a starting point I was reading your definition of the MessageFormat grammar.

Doing so, I noticed some minor things which could be done to simplify the pegjs grammar:

* PegJS grammar:
    * Remove `messageFormatPatternRight` and directly use `messageFormatElement` within `messageFormatPattern`
    * Remove `pluralStyle` and `selectStyle` which were just forwarding to `pluralFormatPattern` respectively `selectFormatPattern`
    * simplified parsing of strings and ids
* automatically compile the grammar in the development version
* I added two new testcases for trailing and leading whitespace in strings

All testcases are passing for messageformat.js and lib/messageformat.dev.js

### DIFF
--- a/lib/message_parser.pegjs
+++ b/lib/message_parser.pegjs
@@ -133,25 +133,19 @@ argStylePattern
   }
 
 string
-  = ws:_ s:(_ chars _)* {
-    var tmp = [];
-    for( var i = 0; i < s.length; ++i ) {
-      for( var j = 0; j < s[ i ].length; ++j ) {
-        tmp.push(s[i][j]);
-      }
-    }
+  = s:(chars/whitespace)* {
     return {
       type: "string",
-      val: ws + tmp.join('')
+      val: s.join('')
     };
   }
 
 // This is a subset to keep code size down
 // More or less, it has to be a single word
 // that doesn't contain punctuation, etc
-id
-  = _ s1:[0-9a-zA-Z$_]s2:[^ \t\n\r,.+={}]* _ {
-    return s1 + (s2 ? s2.join('') : '');
+id "identifier"
+  = _ s:$([0-9a-zA-Z$_]s2:[^ \t\n\r,.+={}]*) _ {
+    return s;
   }
 
 chars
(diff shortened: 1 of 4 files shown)


SlexAxton:
It looks like there are a bunch of failing tests in Travis, are those not from this change set?

vogelsgesang:
Well, the Travis CI tests are not passing since I wrote them for the new grammar and the new syntax tree. For my local tests I replaced `test/common.js` with:

```
MessageFormat = require('../lib/messageformat.dev');
```

instead of

```
MessageFormat = require('../messageformat');
```

But I did not want to commit this change since it might break your testing infrastructure.
What should I do about this?

vogelsgesang:
Or should I apply the changes I applied to `lib/messageformat.dev.js` to `messageformat.js` as well?
Sorry, I guess I just didn't understand your workflow

vogelsgesang:
Ok, after having a look at some other pull-requests I noticed that changes have to be applied to both `messageformat.js` and `lib/messageformat.dev.js`. So I copied my changes over.

The tests in Travis CI on nodejs version 0.10 are passing now. The failing tests under nodejs 0.8 were not from this changeset and are already failing on your current master branch.

SlexAxton:
`messageformat.js` is a generated file. It's mostly a wrapper and adds in the parser. If my memory serves me right, you'd just run `make` in the root. (Might need some npm love first)

vogelsgesang:
`make` worked almost immediately, but I modified the `var mparser` line so your regexp did not match anymore. Therefore I had to adapt my changes of `lib/messageformat.dev.js` first.

In order to avoid broken tests in the individual commits, I rebased my branch.

eemeli:
Thank you @vogelsgesang! I merged this, together with a couple more PEG optimizations & some whitespace cleanup. The previously failing node 0.8 tests were taken out in #88; we now test in node 0.10 & 0.12 and in iojs.

The current custom inlining of the compiled PEG parser will require some thought, and probably replacement.

PR #88, opened 2015-02-11
Add JSDoc comments to source

eemeli:
Working towards better documentation, I added [JSDoc](http://usejsdoc.org/) commentary throughout the source. In addition to being readable _in situ_, it compiles to HTML pretty cleanly.

While at it, I cleaned up the source in a couple of places, most visibly moving `mparser = require('./message_parser')` to be within `MessageFormat._parse`, and removing the `_` prefixes of the runtime functions.

This could be left as is, or we could add `jsdoc` as a dependency, along with a suitable config file and a `make` rule. Here's the minimal `conf.json` I've been using so far:
```json
{
  "source": { "include": [ "lib/messageformat.dev.js" ] },
  "plugins": [ "plugins/markdown" ]
}
```

Regarding the copyright statement, I went for now with "2012-2015 Alex Sexton and other contributors". Is that all right?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /node_modules/**/*
 lib/message_parser.js
+doc/
 
 # Numerous always-ignore extensions
 *.diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
 node_js:
+  - iojs
+  - "0.12"
   - "0.10"
-  - "0.8"
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+Copyright 2012-2015 Alex Sexton, Eemeli Aro, and Contributors
+
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
 "Software"), to deal in the Software without restriction, including
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # 
-# Copyright 2014
+# Copyright 2012-2014 Alex Sexton, Eemeli Aro, and Contributors
 # 
 # Licensed under the MIT License
 # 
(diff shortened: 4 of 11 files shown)


SlexAxton:
All of that sounds good and better. Add your name though. It's half your code now too.

"2012-2015 Alex Sexton, Eemeli Aro, and Contributors"

rxaviers:
Wow, good job

eemeli:
I updated all the copyright statements that I could find to match the same pattern.

The Travis CI build is failing with odd errors in the dependencies for Node 0.8, is that a target we should still care about? I'm pretty sure none of the changes in this commit should be their cause.

SlexAxton:
Kill 0.8, add 0.12 -- and also maybe add some io.js love? http://docs.travis-ci.com/user/languages/javascript-with-nodejs/

SlexAxton:
Ah, 0.12 isn't on travis yet. so just add the io.js stuff, I guess.

SlexAxton:
(we could do 0.11 as a stopgap since the last 0.11 release is invariably the RC for 0.12)

eemeli:
As it turns out, the Travis CI documentation is outdated, and it does support 0.12. :)

SlexAxton:
Boom. This probably eventually warrants killing a lot of the README, and generating a github.io page, but let's take it in steps.

eemeli:
Aye, I agree. Now, should we include jsdoc as a dependency or not? I'd be tempted to put it in as a devDependency, and use it to generate API docs that could get hosted on the aforementioned Github Pages site.

SlexAxton:
Seems like a devDep to me.

eemeli:
I merged this, as we seem to be in agreement. I'll continue on the documentation & readme update when I can.

PR #91, opened 2015-03-07
Use CommonJS

mjackson:
**Note: This is not yet ready for merge because the make-plural dep uses a custom module system similar to messageformat. It should be updated as well.**

This commit removes the custom UMD wrapper and uses the CommonJS
module format. This gives us a few benefits:

- We can use a real require/exports (in the tests too!)
- Compatibility with module loaders like webpack, Browserify

Webpack is used to generate the browser test bundle and a UMD
build in the dist directory.


mjackson:
I need to find some time to submit a PR on the `make-plural` lib as well. For now, I need to drop my dep on `MessageFormat` :( :( and go back to using my ghetto `pluralize` func.

```js
function pluralize(word, n) {
  return n === 1 ? word : word + 's';
}
```

:P

rxaviers:
Have you removed AMD support? Why?

SlexAxton:
I think the idea is that adding AMD would be a build step. I'm fine with
that.

rxaviers:
Gotcha. Thanks

mjackson:
@rxaviers There's a `dist` directory with the built and minified lib in UMD format for those that need it: https://github.com/mjackson/messageformat.js/blob/dee1d368806ee2aa0e4707632c8db7d28f33920e/dist/messageformat.min.js

rxaviers:
The idea of having UMD wrapped files on dist only makes sense to me. Although, can you include the regular (non-minified) UMD?

About committing the dist files... Are you going to update it on every change, every tag? What's the plan?

eemeli:
I like the de-customization of `require()`, but I have a couple of concerns with this PR:

1. It looks like you've updated the `./messageformat.js` compiled file directly, rather than the source `./lib/messageformat.dev.js`. This is by no means the first time a contributor has done that, and it's rather strongly telling me that we need to re-organise the file structure & stop including the compiled output in the repo. We should set up a compiled version in a separate repo for bower (with an actually working `bower.json` file...), and update that on release.

2. I at least am far more comfortable with Browserify rather than Webpack. Is there a particular reason to use the latter here?

3. I don't think this goes far enough. If we're updating the module packaging of the source, at this time I'd much rather see us go for ES6 modules, and use e.g. [Babel](https://babeljs.io/docs/usage/modules/) to transpile them to CommonJS, AMD or anything else.

rxaviers:
> stop including the compiled output in the repo. We should set up a compiled version in a separate repo for bower (with an actually working bower.json file...), and update that on release.

If maintaining a separate repo for distributing code turns out to be overburdened, an alternative approach that could be of help for your consideration is to include the `dist/` files on detached tags. But, not on the `master` branch.

eemeli:
> If maintaining a separate repo for distributing code turns out to be overburdened, an alternative approach that could be of help for your consideration is to include the `dist/` files on detached tags. But, not on the `master` branch.

That actually sounds like a really good idea.

eemeli:
I'm closing this, as I believe the issues raised were addressed by #97 and #98.

Issue #92, opened 2015-03-28
Generated code does not use short-circuit evaluation


vogelsgesang:
The code generated for the messageformat expression
```
{cnt, select,
  unknown {an unknown number of persons}
  other {{cnt, plural,
    one {one person} other {# persons}
  }}
}
```
called with `cnt="unknown` throws the exception:
```
'unknown' isn't a number.
```

This exception is thrown from the runtime.number function ([messageformat.dev.js:193](https://github.com/SlexAxton/messageformat.js/blob/28c66e2eedc9beb1fe91d495bbef086b79c80eb3/lib/messageformat.dev.js#L193)) which is being called for the '#' sign.
First, I was surprised that this function was even called since the "unknown" path should have been taken. After further investigation it turned out that always all branches are evaluated even if only one of them is displayed in the end.

For example for the following template the plural function is evaluated three times, no matter which parameters are passed.
```
{GENDER, select,
  male {He found {NUM_RESULTS, plural,
            one {1 result}
          other {# results}
  }}
  female {She found {NUM_RESULTS, plural,
            one {1 result}
          other {# results}
  }}
  other {They found {NUM_RESULTS, plural,
            one {1 result}
          other {# results}
  }}
}
```

The reason for this behaviour lays in the structure of the generated code. The code for the first example:
```
return select(d.cnt, {
  unknown: "an unknown number of persons",
  other: plural(d.cnt, 0, pluralFuncs.en, {
    one: "one person",
    other: number(d.cnt)+" persons"
  })
});
```
Passing an object with all possible branches to `select` requires the complete evaluation of this object before calling `select`.

What do you think about this behavior? Should I invest some time to think about possible solutions?

eemeli:
You're right, that is a problem. I can think of at least two solutions to the problem, neither of which is clearly better than the other:

First, we could do it "right" and wrap the values of `select`, `selectordinal` and `plural` statements into functions & modify the runtime functions to call these functions. Your example code for instance would then become:

```js
return select(d.cnt, {
  unknown: function() { return "an unknown number of persons"; },
  other: function() { return plural(d.cnt, 0, pluralFuncs.en, {
    one: function() { return "one person"; },
    other: function() { return number(d.cnt)+" persons"; }
  }); }
});
```

That would obviously look much nicer with ES6 arrow notation, but we can't realistically use that yet:

```js
return select(d.cnt, {
  unknown: () => "an unknown number of persons",
[...]

PR #94, opened 2015-04-01
added regression tests for #92

vogelsgesang:
I added 3 regression tests for #92. They are using Messageformat strings like

```
{VAR, select,
     a{correct}
     other {b {VAR, spy}}
     other {incorrect {VAR, spy}}
}
```

where `spy` is a runtime function. The MessageFormat string is called with `VAR = "a"` and the test case fails if the `spy` function was called. I wrote three variations of this test case. One for each of `select`, `plural` and `selectordinal`.

### DIFF
--- a/test/tests.js
+++ b/test/tests.js
@@ -607,6 +607,15 @@ describe( "MessageFormat", function () {
         expect(mfunc({})).to.eql("I am indifferent.");
       });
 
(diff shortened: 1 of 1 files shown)


SlexAxton:
Thanks! I love test-only PRs!

PR #97, opened 2015-04-02
Reorganise file structure: include no compiled output

eemeli:
I'd like to here propose a reorganisation fo the project's file structure, chiefly by removing the compiled `./messageformat.js` from inclusion in the master branch of the repository, and by renaming the core file as `./lib/messageformat.js`, ie. removing the `.dev` that it currently has in the middle of it.

Simultaneously, this allows us to consider the core file directly usable as the node "main" file, and leaves the generated version in the root as the one to use in browser environments.

The main benefit from this change is an increase in clarity regarding the "source" file, which has clearly been experienced by more than one contributor to the project.

This change by itself does make usage in node somewhat slower, due to the live inclusion in `_parse` of the PEG.js parsing of the rules to JavaScript. For that reason, I'm submitting another PR on top of this to fix that.


rxaviers:
Is the plan to include the compiled messageformat.js (e.g., `dist/messageformat.js`) in detached tags?

SlexAxton:
Yea, I think we'll still need to end up committing a built file. I don't really care if it's in detached tags or a 'dist' folder that's always there. We can also just add it in in prepublish. I think the main goal would that it would be harder to find/accidentally edit/accidentally not edit than it currently is, but it'll still be needed for clientside folks.

rxaviers:
:+1:

eemeli:
Yeah, should've mentioned that in the above: detached tags is what I'd suggest for releases. That'll let us continue to have full bower support as well, which practically speaking requires the compiled files.

On a related note, we're also dragging along a `component.json` file that hasn't been updated since #31. Is that a platform we're still supporting? At least the `bower.json` file was completely out of spec until 239c76e, and I've no idea if `component.json` is in similarly bad shape. @pgherveou, would you be willing to take a look at it?

SlexAxton:
Yea, I'm cool with killing if there's no owner.

PR #98, opened 2015-04-02
Replace custom require() hack with Browserify

eemeli:
This PR follows from the changes proposed in PR #97, and would replace the current custom replacement of `require()` in the source with the use of Browserify. For the changes relating to just that, please take a look at just the [second commit](/SlexAxton/messageformat.js/commit/eaf2b20) rather that the whole change.

As it turned out that make-plural's custom loader of CLDR data didn't play nicely with Browserify, I've needed to switch its version to `3.0.0-pr2`, which does work. Eventually make-plural 3.x will have a yet different (ie. simpler) interface for plural functions, at which time I can update messageformat.js as well.

As side benefits of using Browerify, the Bower version loses the external dependency on make-plural, and we gain support for loading as an AMD module.

Also, just to be clear: `npm update` is required by this PR. Furthermore, with this change the PEG.js parser generation is required as a build step even for what was previously the "dev" version of the library, but is now the public-facing version of it too.


SlexAxton:
I'm cool with this.

eemeli:
Excellent!

eemeli:
I got the make-plural output generation done, and integrated it in 22ed78c, which lets us use `require('make-plural/plurals')` directly as `MessageFormat.plurals`, so I could replace the `getPluralFunc()` call in the constructor with three lines of inline code.

Then in c62843b I took out the custom wrapper that was around all of `lib/messageformat.js`; its functionality is now provided by Browserify and double-wrapping doesn't make any sense.

And while doing some other stuff I figured out why the node 0.8 tests were failing in Travis CI, and made them work again with 875596c.

PR #100, opened 2015-04-04
Fixes #99: Code injection using \u005c

### LINKED ISSUE #99: Code injection using \u005c
Arbitrary JS code can be injected into the output of the `Messageformat.compile` function using the escape sequence `\u005c` (unicode code point for a backslash).

Working example:
```
var MessageFormat = require('./lib/messageformat');
var mf = new MessageFormat('en', null);

var injectionString = "\\u005c{VAR} + console.log('Out of the box'); \\}//";
var test = mf.compile(injectionString);

console.log(test({VAR: 'value'}));
```

vogelsgesang:
I used `JSON.stringify` to escape special characters instead of a do-it-yourself solution.

### DIFF
--- a/lib/messageformat.js
+++ b/lib/messageformat.js
@@ -304,13 +304,14 @@ MessageFormat.prototype._precompile = function(ast, data) {
       return '{ ' + r.join(', ') + ' }';
 
     case 'string':
-      tmp = '"' + (ast.val || "").replace(/\n/g, '\\n').replace(/"/g, '\\"') + '"';
+      tmp = JSON.stringify(ast.val || "");
       if ( data.pf_count ) {
         args = [ propname(data.keys[data.pf_count-1], 'd') ];
         if (data.offset[data.pf_count-1]) args.push(data.offset[data.pf_count-1]);
         tmp = tmp.replace(/(^|[^\\])#/g, '$1"+' + 'number(' + args.join(', ') + ')+"');
         tmp = tmp.replace(/^""\+/, '').replace(/\+""$/, '');
       }
+      tmp = tmp.replace(/\\\\#/g, '#');
       return tmp;
 
     default:
--- a/test/tests.js
+++ b/test/tests.js
@@ -441,6 +441,11 @@ describe( "MessageFormat", function () {
         expect((mf.compile('She said "Hello"'))()).to.eql('She said "Hello"');
       });
 
+      it("escapes backslashes (regression test for #99)", function() {
+        var mf = new MessageFormat( 'en' );
+        expect((mf.compile('\\u005c'))()).to.eql('\\');
+      });
+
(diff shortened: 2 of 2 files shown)


lib/messageformat.js
```diff
@@ -304,13 +304,14 @@ MessageFormat.prototype._precompile = function(ast, data) {
       return '{ ' + r.join(', ') + ' }';
 
     case 'string':
-      tmp = '"' + (ast.val || "").replace(/\n/g, '\\n').replace(/"/g, '\\"') + '"';
+      tmp = JSON.stringify(ast.val || "");
       if ( data.pf_count ) {
         args = [ propname(data.keys[data.pf_count-1], 'd') ];
         if (data.offset[data.pf_count-1]) args.push(data.offset[data.pf_count-1]);
-        tmp = tmp.replace(/(^|[^\\])#/g, '$1"+' + 'number(' + args.join(', ') + ')+"');
+        tmp = tmp.replace(/(^|[^\\\\])#/g, '$1"+' + 'number(' + args.join(', ') + ')+"');
```
eemeli: Why this? Doubling the `\` has no effect inside a character class.

vogelsgesang: Yes, you are completely right. I wanted to change line 315 and ended up editing the wrong line.

vogelsgesang:
Thanks for reviewing my commits. You are completely right, I duplicated the backslashes at the wrong line.

Since my thoughts in line 311 were obviously wrong, a short explanation, why I think that two backslashes (after escaping: 4 backslashes) are neccessary in line 314: When parsing `\#` the parser returns a string entity with content `\#`. After applying `JSON.stringify` this yields `\\#`. In the end `#` should be printed and hence two backslashes must be removed.

eemeli:
Of course, the bit I commented on gets replaced by the changes of #102. :)

Using `JSON.stringify` is clearly the better choice than the custom replacement, esp. given the code injection possibility. I also agree that the octothrope parsing fits in well in the PEG.js parser.

vogelsgesang:
Yes, I should have referenced #102. I separated the pull requests in order to keep each one as concise as possible, maybe the wrong decision...

An updated pre-release would be fine, so I can update the bower dependencies of my projects.

eemeli:
That seems like an eminently reasonable request; published now as [0.3.0-1](/SlexAxton/messageformat.js/tree/v0.3.0-1) here and on npm under the `prerelease` tag.

@SlexAxton, for this I packaged the release manually and included the API documentation as well (which needed a little tweaking so the font files wouldn't need to be included). We should consider something like make-plural's [`make release`](/eemeli/make-plural.js/blob/es6/Makefile#L85) to automate it.

...which would help in catching things like the version number in the README that I forgot to update here. Bother.

PR #124, opened 2015-09-28
Refactor out non-cli-specific logic

cletusw:
This enables other tools (like [grunt-messageformat](https://github.com/gushov/grunt-messageformat)) to reuse the globbing/writing code.

### DIFF
--- a/bin/messageformat.js
+++ b/bin/messageformat.js
@@ -61,9 +61,6 @@ var
   },
   options = (function() {
     var o = nopt(knownOpts, shortHands, process.argv, 2);
-    for (var key in defaults) {
-      o[key] = o[key] || defaults[key];
-    }
     if (o.argv.remain) {
       if (o.argv.remain.length >= 1) o.inputdir = o.argv.remain[0];
       if (o.argv.remain.length >= 2) o.output = o.argv.remain[1];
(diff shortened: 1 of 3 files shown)


cletusw:
See [ESLint](http://eslint.org/docs/developer-guide/architecture) for a similar take.

eemeli:
I like the idea, but I'm concerned about a couple of things:
- This puts the option defaults in the runner, while leaving the other parts of option handling in the CLI code. This makes their later management potentially difficult.
- The runner only exposes `runner()`, which forces the use of the `write()` callback. Which internally is rather CLI-specific, with e.g. its check on `options.stdout`

It does makes sense to separate the compiler/runner into its own module, but that module should expose a bit more of its functionality than just the runner, so it can serve potentially wider use cases.

cletusw:
@eemeli Totally agree with you re the write callback. I'll pull that back out of the runner() API.

I'm not so sure about the defaults, though. The benefit of resolving the defaults at this level is that all clients of the API get the same defaults. If we resolve them at the client layer, differences will inevitably appear.

cletusw:
@eemeli Although, if I'm going to abstract out the writing of the files, I should also probably pull out the reading. Should make it better for tools like gulp that only read the files once for a given set of tasks. So basically the purpose of this API is just parseFileSync and managing defaults.

eemeli:
See #138 for a slightly different take on this idea; if accepted, that will effectively supersede this PR.

cletusw:
@eemeli Glad to see that this is getting addressed through #138. Regardless, I'll be moving off of grunt-messageformat in the very near future, so this is no longer something that I'll need. Closing. Thanks!

PR #126, opened 2015-10-02
Package.json - Added browser config for fs

JamesHenry:
So as to avoid webpack warnings being thrown when compiling an app with messageformat in, we need to specify a browser definition for fs.

From the creator of weback to jade:
https://github.com/jadejs/jade/pull/1644

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,5 +51,8 @@
   },
   "scripts": {
     "test": "make test"
+  },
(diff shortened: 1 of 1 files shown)


JamesHenry:
Hey @SlexAxton, would you publishing a new version to npm which includes this commit?

PR #128, opened 2016-01-03
messageformat.js: Provide Bidi Text Direction and Structured Text support to MessageFormat

ashensis:
As per discussion under #539 Message formatting dependant on base text direction I am submitting here the addition of predefined number formatting function enforcing Bidi Text Direction and Structured Text by using UCC. The upgrading qualities are:
 * Prevents intermingling of text contained in placeholders with the rest of message text.
 * Maintains the integrity of Bidi text flowing according to directionality of language script.
 * Enforces 'contextual' Bidi text direction on placeholder content 
 *  i.e. text direction is defined by first strong character of placeholder text - 
 *  'right-to-left' if first strong character is bidi (Hebrew/Arabic) and 'left-to-right' othervise.
 * For reference to Bidi strucctured text see: http://cldr.unicode.org/development/development-process/design-proposals/bidi-handling-of-structured-text
 * For reference to Bidi text direction see: http://w3-03.ibm.com/globalization/page/publish/4353

Please note that this submission constitutes the first step, the next would to be step  is contribution to Globalize (https://github.com/jquery/globalize) with objective to make this Bidi formatter available for 
Globalize.messageFormatter. The intended contribution to Globalize is to:
-  add parameter (object containing formatters) to 'messageFormatter' API and pass it to 'MessageFormat'
      Globalize.prototype.messageFormatter = function( path, fmt ) { 
          ....................
          formatter = new MessageFormat( cldr.locale, pluralGenerator, fmt ).compile( message );
- expose predefined formatters
      Globalize.prototype.messageFormatter.formatters = MessageFormat.formatters;

### DIFF
--- a/lib/messageformat.js
+++ b/lib/messageformat.js
@@ -133,6 +133,48 @@ MessageFormat.formatters = {
   }
 };
 
+/** Predefined number formatting function enforcing Bidi Text Direction and Structured Text by using UCC
+ * Prevents intermingling of text contained in placeholders with the rest of message text.
+ * Maintains the integrity of Bidi text flowing according to directionality of language script.
+ * Enforces 'contextual' Bidi text direction on placeholder content 
+ *  i.e. text direction is defined by first strong character of placeholder text - 
+ *  'right-to-left' if first strong character is bidi (Hebrew/Arabic) and 'left-to-right' othervise.
+ * For reference to Bidi strucctured text see: http://cldr.unicode.org/development/development-process/design-proposals/bidi-handling-of-structured-text
+ * For reference to Bidi text direction see: http://w3-03.ibm.com/globalization/page/publish/4353
+ *  @param {string} - Placeholder text content
+ *  @param {string} - Locale
+ *  @returns {string} - The resultant text fixed by UCC Bidi strong marks
+ *  @examples: (upper case stands for Bidi (Arabic/Hebrew) characters
+ *  > var MessageFormat = require('messageformat');
+ *  > var mf = new MessageFormat('en', null, {"bidiStructuredText": Globalize.messageFormatter.formatters.bidiStructuredText} );
+ *  > var mfunc = mf.compile("{0} >> {1} >> {2}");
+ *  > mfunc([ "first_english_word", "SECOND_ARABIC_WORD", "THIRD_ARABIC_WORD" ])
+ * structured output:
+ *     "first_english_word >> SECOND_ARABIC_WORD  >> THIRD_ARABIC_WORD"
+ * distorted output when MessageFormat gets initialized with no structured text formatter:
+ *     "first_english_word >> THIRD_ARABIC_WORD  << SECOND_ARABIC_WORD"
+ *
+ *  > mf = new MessageFormat('en', null, {"bidiStructuredText": Globalize.messageFormatter.formatters.bidiStructuredText} );
+ *  > mf.compile("FILE {0} NOT FOUND")("c:\filename.EXTENSION");
+ * structured output:
+ *     "FILE c:\filename.EXTENSION NOT FOUND"
+ * distorted output when MessageFormat gets initialized with no structured text formatter:
+ *     "FILE c:\filename.NOT FOUND EXTENSION"
+ *  */
+MessageFormat.formatters.bidiStructuredText = function( text, locale ) {
+	var strongChar = /[A-Za-z\u05d0-\u065f\u066a-\u06ef\u06fa-\u07ff\ufb1d-\ufdff\ufe70-\ufefc]/.exec( text );
+	if ( strongChar && ( strongChar[0] > "z" ) ) { //RLE + text + PDF
+		text = "'\u202B' + " + text + " + '\u202C'";
+	} else { //LRE + text + PDF
+		text = "'\u202A' + " + text + " + '\u202C'";
+	}
+	if ( /he|ar|fa/.test( locale ) ) { //RLM + text + RLM
+		return ( "'\u200F' + " + text + " + '\u200F'" );
+	} else { //LRM + text + LRM 
+		return ( "'\u200E' + " + text + " + '\u200E'" );
+	}
+};
+
 /** Enable or disable support for the default formatters, which require the
  *  `Intl` object. Note that this can't be autodetected, as the environment
  *  in which the formatted text is compiled into Javascript functions is not
@@ -259,7 +301,8 @@ MessageFormat.prototype._precompile = function(ast, data) {
     case 'messageFormatElement':
       data.pf_count = data.pf_count || 0;
       if ( ast.output ) {
-        return propname(ast.argumentIndex, 'd');
+        var ret = propname(ast.argumentIndex, 'd');
+        return this.runtime.fmt.bidiStructuredText ? this.runtime.fmt.bidiStructuredText( ret, this.lc ) : ret;
       }
       else {
         data.keys[data.pf_count] = ast.argumentIndex;
--- a/lib/messageformat.js
+++ b/lib/messageformat.js
@@ -1,3 +1,4 @@
+fstructur
 /** @file messageformat.js - ICU PluralFormat + SelectFormat for JavaScript
  *  @author Alex Sexton - @SlexAxton
  *  @version 0.3.0-0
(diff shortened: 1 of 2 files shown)


eemeli:
Looks like a decent idea, but a few concerns about the implementation:

- You've added your function under `MessageFormat.formatters`. If you'll look at its documentation, you'll see that that object is exposed for use within messageformat syntax. Is that intentional for `bidiStructuredText`? If so, your examples should show a useful example of that.
- The structured output / distrorted output examples in your own code documentation really sound like test cases. You should add those to `test/tests.js`.
- The tests within the function (lines 166 and 171) seem somewhat magical. In particular, is `/he|ar|fa/.test( locale )` an exact way of looking for RTL scripts?
- As is, the function output is rather verbose, with four string concatenations. Can't you reduce or remove those?
- It would be good for your code to match the indentation style of the rest of the file.

tomerm:
Just one observation on the way to look for RTL scripts. Ideally to get natural direction of script (this is what we actually is looking for in this context), one would look at CLDR and trace the connection between language (aka locale) to script and then to direction of script. Unfortunately CLDR data is not available in the context of SleAxton package and thus a simplification was suggested in the submitted PR.

ashensis:
Thank you eemeli for your comments.
However it seems that I didn't understand a few points, could you please elucidate?
1. The first remark, where you said that " .... show a useful example of that".
Sorry, I totally missed your point, could you please to dwell more on this remark?

2. Your first before last remark, "..the function output is rather verbose, with four string concatenations",
again, what is the 'function output' you are speaking of? You see, I am really SLOW.

3.  As for the way of looking for RTL scripts I would definitely reply in confirmatory. Besides the Tomer remark of CLDR date been unavailable in this context, currently CLDR JSON date doesn't contain the
info about the language directionality. Moreover, were it contain such a data, it would be redundant.
Indeed Hebrew, Arabic and Farsi are the only RTL languages (not taking in account such exotic creature as Samaritan, although it recently has been included in Unicode but not present currently and to my believe will not ever be present in CLDR JSON)     
.

eemeli:
Re 1: The members of `MessageFormat.formatters` are available for use directly in messageformat script, so you can have for instance this:
```js
> var MessageFormat = require('messageformat');
> var mf = new MessageFormat('en', null, {number: MessageFormat.number});
> mf.currency = 'EUR';
> var mfunc = mf.compile("The total is {V,number,currency}.");
> mfunc({V:5.5})
"The total is €5.50."
```
As is, your `MessageFormat.formatters.bidiStructuredText` would also be available for that sort of use. If that's intentional, you should include in the documentation an example where that makes sense. If it's not, then you should not have the function be a member of `MessageFormat.formatters`.

Furthermore, why do the examples refer to `Globalize.messageFormatter`? Globalize is by no means the only place where messageformat.js is used.

Re 2: The return of `bidiStructuredText("'example'", 'en')` is currently `"'\u200F' + '\u202A' + 'example' + '\u202C' + '\u200F'"` when it should be `"'\u200F\u202Aexample\u202C\u200F'"` or at least `"'\u200F\u202A' + 'example' + '\u202C\u200F'"`.

Re 3: I'd say it's perfectly fine to include an inlined regexp test for directionality, but it needs to be encoded much more clearly. For example, you'd add in a function `isLocaleLTR(locale)` that returns a boolean, and use that in `bidiStructuredText()`.

Oh, and the JSON version of the CLDR does already contain the LTR/RTL info, e.g. [here for Arabic](https://github.com/unicode-cldr/cldr-misc-full/blob/master/main/ar/layout.json).

eemeli:
Here's an example function, based on CLDR v28 (untested):

```js
function isLocaleRTL(locale) {
  var rtlLanguages = ['ar-.*', 'ar', 'ckb-IR', 'ckb', 'fa-AF', 'fa', 'he', 'ks',
                      'lrc-IQ', 'lrc', 'mzn', 'pa-Arab', 'ps', 'ug', 'ur-IN', 'ur',
                      'uz-Arab', 'yi'];
  return new RegExp('^' + rtlLanguages.join('$|^') + '$').test(locale);
}

**Comment by ashensis** [comment]
Thank you very much eemeli for cooperation.

a) As for Re 1, I already spotted this out, I just took the example from my Globalize tests as is without paying attention.

b) Regarding locale verification I was obviously wrong. However I checked the isLocaleRTL against last 
http://www.unicode.org/repos/cldr-aux/json/26/main/
and found it not enough reliable.
Locales fa* , ks* , pa* , ps*, ur* - have more variations then  checked by isLocaleRTL while locales
ckb, ckb-IR, lrc, mzn, ug, yi - are not found at all
BTW 'yi' is Sichuan and it has ltr script.
So I changed it a bit, relying on http://www.localeplanet.com/icu/ as well.

c) It transpired that my code submission was somewhat premature,
as a matter of fact I can't enforce 'contextual' text direction
since at compillation stage the placeholder parameters are not known yet
and pushing the check on first strong character into compiled function
will make it too awkward, so I decided to put up with BTD support.

I did the requested changes, including test case, could you please have a look.

File: lib/messageformat.js [hunk 1]
```diff
@@ -133,6 +133,36 @@ MessageFormat.formatters = {
...
+ *  > var MessageFormat = require('messageformat');
+ *  > var mf = new MessageFormat('en', null, {"bidiStructuredText": MessageFormat.formatters.bidiStructuredText} );
+ *  > var mfunc = mf.compile("{0} >> {1} >> {2}");
+ *  > mfunc([ "first_english_word", "SECOND_ARABIC_WORD", "THIRD_ARABIC_WORD" ])
+ * structured output:
+ *     "first_english_word >> SECOND_ARABIC_WORD  >> THIRD_ARABIC_WORD"
+ * distorted output when MessageFormat gets initialized with no structured text formatter:
+ *     "first_english_word >> THIRD_ARABIC_WORD  << SECOND_ARABIC_WORD"
+ *  */
+MessageFormat.formatters.bidiStructuredText = function( text, locale ) { //shensis mesage_format
```
eemeli: This method is still in `MessageFormat.formatters`, and there's still no rationalisation for that. Also, there's a spurious comment at the end of this line.

lib/messageformat.js
```diff
@@ -133,6 +133,36 @@ MessageFormat.formatters = {
...
+  var isLocaleRTL = function(locale) {
+    var rtlLanguages = ["ar", "fa", "he", "ks-Arab", 'pa-Arab', "ps",  "ur", "uz-Arab", "ks"];
+    return new RegExp("^" + rtlLanguages.join( "|^" ) + "$").test( locale );
+  };
```
eemeli: Why does the RegExp get a single `$` at the end, verifying that only the last language in the array has its test anchored at the end? As is, `isLocaleRTL('fax')` is `true` while `isLocaleRTL('ksx')` is `false`, for no discernible reason.

test/tests.js
```diff
@@ -612,6 +611,16 @@ describe( "MessageFormat", function () {
...
+      it("should use bidiStructuredText MessageFormat formatter", function () {
+        var mf = new MessageFormat('en', null, {"bidiStructuredText": MessageFormat.formatters.bidiStructuredText} );
+        var mfunc = mf.compile("{0} >> {1}");		
+        expect(mfunc([ "first_english_word", "SECOND_ARABIC_WORD" ])).to.equal('\u200efirst_english_word\u200e >> \u200eSECOND_ARABIC_WORD\u200e');
+
+        mf = new MessageFormat('ar-EG', null, {"bidiStructuredText": MessageFormat.formatters.bidiStructuredText} );
+        var mfunc = mf.compile("{0} >> {1}");		
+        expect(mfunc([ "first_english_word", "SECOND_ARABIC_WORD" ])).to.equal('\u200ffirst_english_word\u200f >> \u200fSECOND_ARABIC_WORD\u200f');
+      });
+
```
eemeli: As we're dealing with Unicode, there's no reason why `SECOND_ARABIC_WORD` isn't here a word in actual Arabic.

lib/messageformat.js
```diff
@@ -133,6 +133,36 @@ MessageFormat.formatters = {
...
+/** Predefined number formatting function enforcing Bidi Structured Text by using UCC
+ * Prevents intermingling of text contained in placeholders with the rest of message text.
+ * Maintains the integrity of Bidi text flowing according to directionality of language script.
+ * For reference to Bidi structured text see: http://cldr.unicode.org/development/development-process/design-proposals/bidi-handling-of-structured-text
+ *  @param {string} - Placeholder text content
+ *  @param {string} - Locale
+ *  @returns {string} - The resultant text fixed by UCC Bidi strong marks
+ *  @examples: (upper case stands for Bidi (Arabic/Hebrew) characters, output is shown as displayed in browser page
+ *  > var MessageFormat = require('messageformat');
+ *  > var mf = new MessageFormat('en', null, {"bidiStructuredText": MessageFormat.formatters.bidiStructuredText} );
```
eemeli: Is there a reason why the `bidiStructuredText` function needs to be user-modifiable rather than provided directly by the library? I'm essentially looking for a reason why the incantation can't be `var mf = new MessageFormat('en', null, {bidiStructuredText: true})`.

ashensis:
Sorry for residue comment, it will be removed.
Now regarding your last comment. 
There are utility functions under MessageFormat.prototype.runtime. These are used via notation like MessageFormat.number.
I placed my formatter under MessageFormat.formatters and use it appropriately.
So do you propose to move my formatter under MessageFormat.prototype.runtime in order to call it MessageFormat.bidiStructuredText?
Then could you please explain me the difference between those defined in  MessageFormat propo and others defined under MessageFormat.formatters?

eemeli:
This is the comment in the code for `MessageFormat.prototype.runtime`:
```js
/** A set of utility functions that are called by the compiled Javascript
 *  functions, these are included locally in the output of {@link
 *  MessageFormat#compile compile()}.
 *
 *  @namespace
 *  @memberof MessageFormat  */
```

This is the comment in the code for `MessageFormat.formatters`, note in particular the examples:
```js
/** Default number formatting functions in the style of ICU's
 *  {@link http://icu-project.org/apiref/icu4j/com/ibm/icu/text/MessageFormat.html simpleArg syntax}
 *  implemented using the
 *  {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl Intl}
 *  object defined by ECMA-402.
 *
 *  **Note**: Intl is not defined in default Node until 0.11.15 / 0.12.0, so
 *  earlier versions require a {@link https://www.npmjs.com/package/intl polyfill}.
 *  Therefore {@link MessageFormat.withIntlSupport} needs to be true for these
 *  functions to be available for inclusion in the output.
 *
 *  @see MessageFormat#setIntlSupport
 *
 *  @namespace
 *  @memberof MessageFormat
 *  @property {function} number - Represent a number as an integer, percent or currency value
 *  @property {function} date - Represent a date as a full/long/default/short string
 *  @property {function} time - Represent a time as a full/long/default/short string
 *
 *  @example
 *  > var MessageFormat = require('messageformat');
 *  > var mf = (new MessageFormat('en')).setIntlSupport(true);
 *  > mf.currency = 'EUR';
 *  > var mfunc = mf.compile("The total is {V,number,currency}.");
 *  > mfunc({V:5.5})
 *  "The total is €5.50."
 *
 *  @example
 *  > var MessageFormat = require('messageformat');
 *  > var mf = new MessageFormat('en', null, {number: MessageFormat.number});
 *  > mf.currency = 'EUR';
 *  > var mfunc = mf.compile("The total is {V,number,currency}.");
 *  > mfunc({V:5.5})
 *  "The total is €5.50."  */
```

Now, consider your new `bidiStructuredText`. Where and when is it called from? How visible does it need to be? In particular, is there any reason to include it in the compiled output that `MessageFormat.prototype.compile()` provides?

The answer, I suspect, is that it doesn't need any visibility at all except for being able to be called from within `MessageFormat.prototype.compile()`. And that means it should not be visible anywhere else. And that means it's really a utility function like `propname()` and belongs next to that.

If you don't understand why, please study the code some more and ask specific questions about what you don't understand.

lib/messageformat.js
ashensis: I am relying on input locale name to be meaningful. This being said, I test positively all locales started with array members listed in rtlLanguages (like "ar-EG" , "ks-Arab" or "pa-Arab") except "ks" which I test for exact match to exclude "ksb".
In all those are RTL locales that I do  recognize:
ar*
he
fa*
=====
ks
ks-Arab
ks-Arab
 but not
ksb, ksf, ksh
=====
pa-Arab
pa-Arab-PK
 but not
pa
pa-Guru
pa-Guru-IN
=====
ps
ps-AF

ur
ur-IN
ur-PK

uz-Arab
uz-Arab-AF
    but not
uz
uz-Cyrl	
uz-Cyrl_UZ	
uz-Latn 	
uz-Latn_UZ

ashensis:
Sorry, i already mentioned that i am SLOW, now got it. Will refurbish

ashensis:
So what do you think, how is your proposal fits with my intentions?

eemeli:
Fundamentally, the problem is that with your current code the output can be broken because `bidiStructuredText` is available for calling through markup, when it obviously should not be:

```js
var mf = new MessageFormat('en', null, {"bidiStructuredText": MessageFormat.formatters.bidiStructuredText} );
var mfunc = mf.compile("a{X,bidiStructuredText}");
mfunc({X:"x"})
// "a'\u200e' + x + '\u200e'"
```

It is **not** a formatting function, and it is **not** a runtime function. It's a utility function. Please fix, and resubmit with a new PR.

ashensis:
Ok eemeli, I only would like to verify what should be the approach, form your previous remarks I surmise that you expect this mode of calling: 
mf = new MessageFormat('en', null, {bidiStructuredText: true});
while MessageFormat will invoke the internal (library) function based on this parameter . Is my understanding correct?

eemeli:
Yes.

ashensis:
Thank you very much, i will open another PR. (Ironically this solution proposed by you I implemented at first but later reworked it to use the customer defined formatter in order to be less invasive)

tomerm:
@eemeli  Just to better understand the process with you kind permission. @ashensis is planning to address all your comments and submit a fixed version of his code suggestion. How should we interpret the action of closing this specific PR ? Is it OK from your perspective to submit a new PR on the same subject / problem with code addressing all your comments ? Can we submit the code addressing all your comments still as part of this PR ?

eemeli:
Either re-opening this or making a new one are fine options, but please do rebase it to a single change; this series of commits is getting rather dirty.

tomerm:
@eemeli thanks ! @ashensis submitted a new PR: 
https://github.com/SlexAxton/messageformat.js/pull/129

PR #132, opened 2016-02-10
Added support for compiling more deeply nested namespaces

hadrianoliver:
I'm working on an system that uses more levels of namespaces than are currently supported by the compile function. This change adds support for any arbitrary depth of namespace nesting by using a recursive function to handle them instead of the current nested for loops.

### Example:
```
var messages = {
  namespace1: {
    namespace2: {
      message: "Compile can't handle this without this change!"
    }
  }
}

var MessageFormat = require('messageformat');
var messageformat = new MessageFormat();
var compiled = messageformat.compile(messages);
```

Results in the error:

> Warning: Parser error with `namespace2` in `namespace1`: TypeError: Object #<Object> has no method 'charCodeAt' Use --force to continue.

### DIFF
--- a/lib/messageformat.js
+++ b/lib/messageformat.js
@@ -490,6 +490,18 @@ MessageFormat.prototype.compile = function ( messages, opt ) {
         for (var i = 0; i < level; ++i) indent += '  ';
         for (var k in r) o.push('\n' + indent + '  ' + propname(k) + ': ' + stringify(r[k], level + 1));
         return '{' + o.join(',') + '\n' + indent + '}';
+      },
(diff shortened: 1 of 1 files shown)


eemeli:
This looks pretty good. If you could add a [test case](/SlexAxton/messageformat.js/blob/master/test/tests.js) for the new functionality, I'll accept the PR.

hadrianoliver:
On reflection I don't believe this change would be good, even though it works as I had intended. I'd rather change the system I'm working on to behave in a more 'usual' manner and make the behaviour of MessageFormat::compile() consistent with the output from the build-time compilation. When I get this working in a way I'm happy with I'll make another pull request with tests.

PR #133, opened 2016-02-11
Enclose JavaScript reserved words used as keys in quotes

hadrianoliver:
When compiled objects use JavaScript reserved words as keys they are not being enclosed in quotes which causes errors in IE <= 8. This change corrects the problem.

### Example:
In this example the JavaScript reserved word `default` is used as a key.
```
var messages = {
  default: "IE <= 8 can't load the compiled script because 'default' is a reserved word in JavaScript"
}

var MessageFormat = require('messageformat');
var messageformat = new MessageFormat();
var compiled = messageformat.compile(messages);

compiled.toString();
```
### Output:
```
function anonymous() {
var number = function (value, offset) {
  if (isNaN(value)) throw new Error("'" + value + "' isn't a number.");
  return value - (offset || 0);
};
var plural = function (value, offset, lcfunc, data, isOrdinal) {
  if ({}.hasOwnProperty.call(data, value)) return data[value]();
  if (offset) value -= offset;
  var key = lcfunc(value, isOrdinal);
  if (key in data) return data[key]();
  return data.other();
};
var select = function (value, data) {
  if ({}.hasOwnProperty.call(data, value)) return data[value]();
  return data.other()
};
var pluralFuncs = {
  en: function (n, ord) {
    var s = String(n).split('.'), v0 = !s[1], t0 = Number(s[0]) == n,
        n10 = t0 && s[0].slice(-1), n100 = t0 && s[0].slice(-2);
    if (ord) return (n10 == 1 && n100 != 11) ? 'one'
        : (n10 == 2 && n100 != 12) ? 'two'
        : (n10 == 3 && n100 != 13) ? 'few'
        : 'other';
    return (n == 1 && v0) ? 'one' : 'other';
  }
};
var fmt = {};

return {
  default: function(d) { return "IE <= 8 can't load the compiled script because 'default' is a reserved word in JavaScript"; }
}
}

```

### DIFF
--- a/lib/messageformat.js
+++ b/lib/messageformat.js
@@ -8,18 +8,17 @@
 /** Utility function for quoting an Object's key value iff required
  *  @private  */
 function propname(key, obj) {
-  var name = '';
-  try {
-    if (/^[A-Z_$][0-9A-Z_$]*$/i.test(key) && new Function(key, '')) {
-      name = obj ? obj + '.' + key : key;
-    }
-  } catch (e) {
-  } finally {
-    if (!name) {
-      var jkey = JSON.stringify(key);
-      name = obj ? obj + '[' + jkey + ']' : jkey;
-    }
-    return name;
+  /* Quote the key if it contains invalid characters or is an
+   * ECMAScript 3rd Edition reserved word.
+   */
+  if (/^[A-Z_$][0-9A-Z_$]*$/i.test(key) &&
+     ['break', 'continue', 'delete', 'else', 'for', 'function', 'if', 'in', 'new',
+      'return', 'this', 'typeof', 'var', 'void', 'while', 'with', 'case', 'catch',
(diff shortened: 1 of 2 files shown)


eemeli:
I'm not sure I like the idea of using `new Function()` just to test if a word is reserved. What's happening here with IE8 is that it's following the ES3 standard, which prohibits the use of reserved words with dot notation. ES5 changed that, which explains why e.g. `default` is now ok as a key. Furthermore, it [looks like](/jshint/jshint/issues/674) the future reserved words were never checked, which limits the set to the following ES3 reserved words:

```
break case catch continue default delete do else finally for function if in
instanceof new return switch this throw try typeof var void while with
```

So rather than compiling a function just to get an exception, why don't we check for a match to that list?

hadrianoliver:
Agreed. Have changed to check against a list of reserved words. Looks much nicer now.

PR #135, opened 2016-02-20
Remove component.json

eemeli:
We should deprecate support for component, as it [is no longer maintained](/componentjs/component/issues/639). Also, I at least wasn't able to find messageformat on [component.github.io](http://component.github.io/), and component.io appears to longer be operational.


SlexAxton:
Yep.

Issue #110, opened 2015-06-29
Ship minified version


kdex:
The `Angular` module `angular-translate-interpolation-messageformat` uses `messageformat.js` in order to support pluralization and other features.

As you might expect, this means that many browsers are in need of `messageformat.js`. At the moment, however, all that's shipped is an overly verbose (that is, unminified) version that is not suitable for the web at **47.7 KiB** (this is a *lot*).

Most package maintainers therefore minify their scripts for browsers so that users can use tools such as `wiredep` to automatically hook up all bower modules into their markup and tools such as `usemin` to bundle them into one big file. This has one big advantage: Users can have their dependencies minified in a file without having to minify even one line themselves. Build time will therefore dramatically decrease for everyone.

With a potential **21.8 KiB**, messageformat.js could highly benefit from shipping a minified version.

So, erm: Pretty please? :)

SlexAxton:
Thank you for the thoughtful explanation.

I'm not 100% opposed to this, but if it's performance you're after, offering a minified version seems like it's promoting a bad practice.

If you want the best performance, you'll already need a process for bundling scripts and minifying them. I think offering a minified version has fallen out of favor because it implies that someone would include your script as a separate resource.

When the script isn't minified in dev mode, the errors are also easier to understand, and source maps can work nicely after your build process has run, even in production.

Is there a scenario, where providing minified version is going to promote the best practice, that I'm missing?

SlexAxton:
(Also, during the build process, people should be precompiling their templates and then only including the vastly smaller messageformat.runtime.js - bundling that - and then minifying all of it.)

kdex:
Well, alongside minified versions, most bower package maintainers (compare [bower-angular](https://github.com/angular/bower-angular), [jQuery](https://github.com/components/jquery), [underscore](https://github.com/jashkenas/underscore) and others) do also ship source maps themselves.

Although I can follow that shipping minified versions doesn't seem like "good practice" at first (since it feels equally bad as committing binary formats to a repository), there are scenarios where this is beneficial:

Consider the use of `messageformat.js` from a `bower` point of view. Bower specifies a `main` property in `bower.json`. `main` takes an array of strings representing relative paths to certain files in your bower package.

[...]

PR #137, opened 2016-02-24
Ship minified version

eemeli:
This implements #110, including the minified input in bower releases, along with a source map.


SlexAxton:
:+1: looks good pending tests.

Issue #105, opened 2015-04-30
# signs within SelectFormat


vogelsgesang:
With the current implementation the messageformat string

```
{VAR1, plural,
  one {1}
  other {{VAR2, select,
    other {#}
  }}
}
```

is equivalent to

```
{VAR1, plural,
  one {1}
  other {{VAR2, select,
    other {{VAR2}}
  }}
}
```

I expected SelectFormat not to interfere with the `#` since [the standard for SelectFormat](http://icu-project.org/apiref/icu4j/com/ibm/icu/text/SelectFormat.html) doesn't even mention the `#` sign.

If I am reading [the standard for PluralFormat](http://icu-project.org/apiref/icu4j/com/ibm/icu/text/PluralFormat.html) correctly
```
An unquoted pound sign (#) in the selected sub-message
itself (i.e., outside of arguments nested in the sub-message)
is replaced [...]
```
the `#` in my string should simply yield a literal `#`.

Which behavior do you consider to be correct?

eemeli:
You're right, our quoting doesn't match the standard. Here's the [official documention](http://icu-project.org/apiref/icu4j/com/ibm/icu/text/MessageFormat.html)'s take on them (ignore the choiceStyle bit, we don't implement that):

> * messageText can contain quoted literal strings including syntax characters. A quoted literal string begins with an ASCII apostrophe and a syntax character (usually a {curly brace}) and continues until the next single apostrophe. A double ASCII apostrohpe inside or outside of a quoted string represents one literal apostrophe.
* Quotable syntax characters are the {curly braces} in all messageText parts, plus the '#' sign in a messageText immediately inside a pluralStyle, and the '|' symbol in a messageText immediately inside a choiceStyle.

And this is from our readme:

> messageformat.js tries to a good job of being tolerant of as much as possible, but some characters, like the ones used the actual MessageFormat spec itself, must be escaped to be a part of your string.

> For `{`, `}` and `#` (only inside of a select value) literals, just escape them with a backslash. (If you are in a JS string, you'll need to escape the escape backslash so it'll look like two).

[...]

PR #138, opened 2016-02-24
Refactor everything

### LINKED ISSUE #134: Error when using with at least 2 languages
Each new `MessageFormat` instance creation totally removes (see it [here](https://github.com/SlexAxton/messageformat.js/blob/master/lib/messageformat.js#L66)) all plural functions from `MessageFormat.prototype.runtime.pluralFuncs` object that is shared accross all MessageFormat instances. Why do you do this?

You should remove this line or make `runtime`/`runtime.pluralFuncs` per-instance, not global.

This bug in Plunker:

1. Go to http://plnkr.co/edit/TZQ1VR2C0JNR1vaJNT4T?p=preview
2. Open console and see `Uncaught TypeError: lcfunc is not a function`

eemeli:
So, I may have ended up going down a bit of a rabbit hole when I fixed #134 while thinking of #132, and put together this little proposed set of changes. It includes changes to, well, nearly everything except for the parser. The [source code](/eemeli/messageformat.js/blob/simpler-output/lib/messageformat.js) documents itself pretty well, but here's an overview of the changes:

#### Constructor
Takes in a single `locale` parameter, which can be a string, array or object; the last form effectively includes the functionality of the previous second parameter `pluralFunc`. The semantics of `locale` are also a little different, as the concept of fallback locales is dropped, and instead having multiple entries creates a fully multi-lingual `MessageFormat` object. The formatters are dropped from the constructor; their addition is now with a chainable `addFormatters` method. They're also much better documented.

#### `compile()`
With object input, the output is no longer a function, but an object with a similar structure as the input, with each messageformatted string replaced by a corresponding function. The top-level object has a [`toString` method](/eemeli/messageformat.js/blob/simpler-output/lib/messageformat.js#L634) that works a bit like previously, except better -- it defaults to an UMD wrapper, and can output e.g. ES6 module code. When used with multiple locales, the locale is auto-detected from the object keys. The second optional parameter `locale` can be used to force a single locale, or to select one if the constructor was originally called with no parameters.

#### Runtime
The included runtime methods are flattened, and only included if actually used. See the updated [example](/eemeli/messageformat.js/blob/simpler-output/example/i18n.js) for the new default output, and note that it's missing the `select` function as that's not included in any of the compiled functions. The  flattening does mean that some locale names need to be fiddled with to make them valid non-reserved JS function names.

#### bin/messageformat.js
Effectively completely rewritten and simplified, given the new `compile().toString()` output and options.

----

Altogether, this is a whole pile of breaking changes, and will need a new major version. Perhaps it'd be time to consider a 1.0 release?


SlexAxton:
Lol wow, nice work. I might need a few days to go through this. That cool?

SlexAxton:
(and I'm sure you see tests are failing)

eemeli:
Taking a little time to review is absolutely fine. :)

Found the test-failure-generator. It turns out the node `fs.write()` call has a [form](https://nodejs.org/api/fs.html#fs_fs_write_fd_data_position_encoding_callback) I'm using in tests that was only introduced in node 0.12, and we're now failing with 0.8 and 0.10. Will fix.

eemeli:
Any thoughts yet, @SlexAxton?

eemeli:
I spent another day in the code, and started by simplifying the parser. Then I did some other updates:
- The previous `lib/messageformat.js` has now effectively been split into `lib/index.js` and `lib/compiler.js`, with the latter being a collection of utilities and an evolved/simplified form of the previous `MessageFormat.prototype._precompile` method.
- The output no longer has wrapping functions for each plural/select/selectordinal case, thereby reducing output complexity significantly, but adding the requirement that formatting functions should be pure and not have side effects.
- A new method `setStrictNumberSign()` controls the treatment of the `#` character inside cases; the default is now to seek the nearest surrounding plural or selectordinal statement's variable. Setting this true will pass all `#` through unmodified if not directly in a plural or selectordinal statement. The previous behaviour was to try and us the value of the immediately surrounding case, and failing if e.g. it's a select with a non-numerical value. (Issue #105)
- Plural and selectordinal keys are now required to be valid keys for the current locale. (PR #96)
- Whitespace is preserved inside cases. (Issue #144)
- The compiled `messageformat.js` is included in the npm-published version (Issue #142)

SlexAxton:
You're a machine. The only thing that sticks out to me is:

> Plural and selectordinal keys are now required to be valid keys for the current locale. (PR #96)

This would intoduce a runtime error (I know it's compile time, but lots of folks don't precompile) for some folks that goes against the standard implementations afaict. 

I read through the old issue. Are we doing any sort of more silent or warning type thing? Or does it just fail across the board?

Yay whitespace preservation too! :)

SlexAxton:
My other thought:

The format-js guys technically pulled out our old peg parser and changed some things but otherwise were kind of very similar. I very seriously doubt we could use their peg as it is, but since there's is an external dependency, maybe if we forked it, made our necessary changes and used it as a dependency it'd increase the chances of our implementations at least converging at the AST level.

Does that sound crazy?

eemeli:
Regarding the plural/selectordinal keys, the spec is actually a bit weird on this, as it doesn't limit the keys, but does expect and require the input for the matching to be a number. Effectively, it's messageformat.js that's been incorrect here because we've not required the input to be a number. I checked a few of the other implementations, and it may just have been us that worked this way.

I have now made the generated error more descriptive, and added a method `disablePluralKeyChecks()` to return to the previous functionality. Another option would be to keep the checks, but `console.warn` instead of throwing an error.

While doing that, I also moved the make-plural `require()` statements back inside `getPluralFunc()` -- which also needed to get a bit more complex for the checks to be optional in the parser. This way, if you're bringing in plural functions yourself, make-plural is never called.

Regarding the parser, I presume you mean [this](/yahoo/intl-messageformat-parser/blob/master/src/parser.pegjs)? I did take a look at it while refactoring our parser, but ended up doing things a bit differently; the version I [ended up with](/eemeli/messageformat.js/blob/simpler-output/lib/parser.pegjs) is IMHO much simpler. Changing the yahoo/intl-messageformat [compiler](/yahoo/intl-messageformat/blob/master/src/compiler.js) doesn't look trivial, but I agree that there may be use cases for our parser to be used by itself. I'll have a try at extracting it from this repo, and see how it looks.

Given the scope of the changes in this PR, it might make sense to push out a `v1.0.0-rc.1` prerelease version from it, and see what breaks in real use cases. Would that be ok?

SlexAxton:
Go for it. Looking solid.

eemeli:
That was decently painless. I extracted the full history of edits to the parser, and constructed a new repo around that at [messageformat/parser](/messageformat/parser). That's now published to npm with a "latest" release of `v1.0.0-rc.1`, and can be installed with:
```sh
npm install messageformat-parser
```

I've subsequently removed the parser code from messageformat.js, messed around a bit, and got it released on bower and npm:
```sh
bower install messageformat#1.0.0-rc.2
npm install messageformat@prerelease
```

I think despite my fixes, the bower release may still be missing the main jsdoc style file. I'll look at it later. I've left this branch on my own account for now, as github won't let the base branch of an existing PR to be changed. There is, however, the github prerelease tag [`v1.0.0-rc.2`](/SlexAxton/messageformat.js/releases/tag/v1.0.0-rc.2) that bower uses.

cletusw:
@eemeli @SlexAxton Any ETA on this? Anecdotally, I've tried 1.0.0-rc.2 in a very large project and it seems to be working fine, but I'm hesitant to push it to production in case there are known issues that I've just missed.

SlexAxton:
I feel pretty confident that the RC2 is pretty close to what will be released. Usually it's docs and stuff that are holding back releases, the code seems to be pretty solid.

cletusw:
@SlexAxton @eemeli Sorry to keep pestering -- just take it as meaning somebody really appreciates your work :)

What's left to be done for the 1.0.0 release? Anything I can help with?

eemeli:
@cletusw, the main reason for slow progress is lack of time and focus on other projects. I finally got around to addressing the issues that had popped up with RC2, and pushed out [RC3](/SlexAxton/messageformat.js/releases/tag/v1.0.0-rc.3). We can probably re-release it as `1.0.0` shortly, unless any further issues come up with it.

cletusw:
I totally understand. Thanks again for the amazing work!

eemeli:
I merged the branch to `SlexAxton/master`, as I probably should've done a while back. This should help e.g. @cletusw base his octothorpe error PR properly.

PR #151, opened 2016-07-15
Update watchr dep

balupton:
A dep of watchr is causing installs to fail, as per https://github.com/bevry/typechecker/issues/13

I've released v2.5.0 of watchr, so this will need to be updated. Watchr follows semver, so changed ~ to ^


SlexAxton:
Great. Thanks!

PR #156, opened 2016-08-01
Throw fewer runtime errors for `#`

eemeli:
This relaxes the default runtime treatment of `#`, and expands the meaning of  `MessageFormat#setStrictNumberSign(true)` to return to the previous, stricter treatment.

Before, we had this:
```js
var mf = new MessageFormat('en');
var msg = '{X, plural, one{the thing} other{# things}}';
mf.compile(msg)({ X: 3 });
  // '3 things'
mf.compile(msg)({ X: 'cool' });
  // Error: 'cool' from argument 'X' isn't a number.
```

With this PR, we get this:
```js
var mf = new MessageFormat('en');
var msg = '{X, plural, one{the thing} other{# things}}';
mf.compile(msg)({ X: 3 });
  // '3 things'
mf.compile(msg)({ X: 'cool' });
  // 'cool things'
mf.compile(msg)({});
  // 'undefined things'
mf.setStrictNumberSign(true);
mf.compile(msg)({ X: 'cool' });
  // Error: Argument `X` has non-numerical value "cool".
```

To get that to work, I needed to make the runtime depend on the MessageFormat class, which was easier to do by separating Runtime to be more clearly its own class, defined in `lib/runtime.js`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ CHK=${GREEN} ✓${STOP}
 ERR=${RED} ✖${STOP}
 
 BIN=./node_modules/.bin
-SRC=lib/index.js lib/compiler.js
+SRC=lib/index.js lib/compiler.js lib/runtime.js
 
 .PHONY: all test test-browser doc release clean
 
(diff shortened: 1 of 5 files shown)


SlexAxton:
I don't think anyone is going to turn on `setStrictNumber` in real life, but I'm roughly fine with that. I'd be more comfortable if we added that option to the docs before we merged, but am cool either way.

eemeli:
That may well be, but I at least am happier with us supporting standard-compliance at least optionally. At least the API documentation will get updated from the JSDoc comments, but the public site could indeed do with an update to match the 1.0 release.

I'll merge this and see if I could get that finally done as well.

cletusw:
@eemeli Feel free to ping me on your docs changes if you'd like a second pair of eyes on the changes.

PR #164, opened 2016-11-02
Fix typos in doc comments

nkovacs:
iff -> if


eemeli:
That was actually not a typo; see [iff](https://en.wikipedia.org/wiki/Iff). However, just using "if" here isn't wrong either, and is probably clearer.

PR #165, opened 2016-11-04
extra features for the JS module generator

pavelstudeny:
1. fix: use path.sep as a path separator instead of '/' to make it work better on Windows

2. option to use a module.exports structure as-is in the JSON input file, to avoid a structure like
`module.exports = {
  "c:": {
    Users: {
      username: {
        src: {
          locale: {
            en: {
              messages: {
`

3. filter keys in the JSON input file, as e.g. some translation services use a "description" key that should not get into translation functions

### DIFF
--- a/bin/messageformat.js
+++ b/bin/messageformat.js
@@ -32,7 +32,7 @@ if (options.help || inputFiles.length === 0) {
 } else {
   var locale = options.locale ? options.locale.join(',').split(/[ ,]+/) : null;
   if (inputFiles.length === 0) inputFiles = [ process.cwd() ];
-  var input = readInput(inputFiles, '.json', '/');
+  var input = readInput(inputFiles, '.json', path.sep);
   var ns = options.namespace || 'module.exports';
   var mf = new MessageFormat(locale);
   if (options['disable-plural-key-checks']) mf.disablePluralKeyChecks();
--- a/bin/messageformat.js
+++ b/bin/messageformat.js
@@ -15,13 +15,15 @@ var fs = require('fs'),
       help: Boolean,
       locale: [String, Array],
       namespace: String,
-      'disable-plural-key-checks': Boolean
+      'disable-plural-key-checks': Boolean,
+      'flat': Boolean
     },
     shortHands = {
       h: ['--help'],
       l: ['--locale'],
       n: ['--namespace'],
-      p: ['--disable-plural-key-checks']
+      p: ['--disable-plural-key-checks'],
+      f: ['--flat']
     },
     options = nopt(knownOpts, shortHands, process.argv, 2),
     inputFiles = options.argv.remain.map(function(fn) { return path.resolve(fn); });
@@ -32,7 +34,7 @@ if (options.help || inputFiles.length === 0) {
 } else {
   var locale = options.locale ? options.locale.join(',').split(/[ ,]+/) : null;
   if (inputFiles.length === 0) inputFiles = [ process.cwd() ];
-  var input = readInput(inputFiles, '.json', path.sep);
+  var input = readInput(inputFiles, '.json', options.flat ? null : path.sep);
   var ns = options.namespace || 'module.exports';
   var mf = new MessageFormat(locale);
   if (options['disable-plural-key-checks']) mf.disablePluralKeyChecks();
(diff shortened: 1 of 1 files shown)


eemeli:
This looks interesting, but I'm not sure I understand what exactly the problems are that are being solved here. Could you provide an example or two of how these new options would work, and what sort of new use cases they would enable?

pavelstudeny:
Sure. The use case here, that is more or less a standard for several environments and translation services, is a file layout and structure like this:

```
<locale directory>
+- de
|    +- messages.json
|         {
|           "key1" : {
|             "message": "translated string",
|             "description": "extra explanation for a translator"
|           },
|           "key2": ...
+- en
|    +- messages.json
... 
```

Naturally, because each language has different plural rules etc., each goes to a separate file, like de.js, en.js, ...

With the curremt status, the exports of the js file looks like

```
module.exports = {
  <locale directory>: {
    de: {
      messages: {
        key1: {
          message: function(d) { return "translated string"; },
          description: function (d) { return "extra explanation for a translator"; }
        },
...
```

The `--regex` option allows to filter out the `description`, that doesn't make sense here. The `--flat` option disregards the directory structure and only puts the in-JSON structure into the js file. With these options combined, the output looks like

```
module.exports = {
        key1: {
          message: function(d) { return "translated string"; },
        },
...
```

eemeli:
Okay, you're right that the structure is currently needlessly complex, but there are a couple of issues in your solutions. The `--flat` option, for instance, will break for cases where there are multiple input files, and doesn't take into account potential complexity at the root of the JSON file. This got stuck in my head earlier today, and I came up with [this](../compare/master...eemeli:bin-simplify) alternative solution to the same problem, and its extension of having useless complexity in the middle of the structure as well. Here are some of the results with these changes:

```
messageformat.js$ messageformat example/en/colors.json

module.exports = { red: function(d) { return "red"; }, ... }
```

```
messageformat.js$ messageformat example/*/colors.json

module.exports = {
  en: {
    colors: { red: function(d) { return "red"; }, ... }
  },
  fr: {
    colors: { red: function(d) { return "rouge"; }, ... }
  }
}
```

```
messageformat.js$ messageformat --simplify example/*/colors.json

module.exports = {
  en: { red: function(d) { return "red"; }, ... },
  fr: { red: function(d) { return "rouge"; }, ... }
}
```

Would this be suitable for your purposes? One concern here is that there's a change in the shape of the output for the default case, as the root key gets dropped. Also, for the degenerate case of having a structure that only contains one value, it ends up not generating the namespace correctly, on account of the input to `compile()` not looking like an object.

Regarding the `--regex` option, is there really a use case for making it that generic, and requiring a regexp to match for the full path? Would something like `--ignore-keys=blah,description` be clearer and sufficient for actual use?

pavelstudeny:
As a side note, the example of mixing English and French into a single output file should not be regularly used, because that would make both translations to use the same locale rules, such as number and plural definitions.

However, if a translation for a single language is split into several files, you could indeed run into problems with duplicate keys. That's why I'm suggesting this as an option instead of altering the current behavior - in a single input -> single output scenario, you may want to use --flat option, in a multiple inputs -> single output behavior, you may not.

I agree that the --regex option is quite generic, but I was afraid about unexpected surprises when using a tree-ish structure like
```
description: {
  msg: "...",
  description: "..."
}
```
A regular expression allows to be very specific about what to include or ignore. But your counter-proposal should also suffice in most cases.

eemeli:
Actually, compiling multiple languages into a single file is perfectly fine, as long as the appropriate locales are included. Take a look for instance at the [example output](../blob/v1.0.2/example/i18n.js), which is generated with the following command:

```
messageformat --locale=en,fr --namespace=i18n example/ > example/i18n.js
```

If you look at its lines 42 and 56, you'll see that the correct locale has been auto-detected by matching parts of the path against the available locales. This is documented [here](https://messageformat.github.io/messageformat.js/doc/Compiler.html#compile).

Regarding `--ignore-keys`, an entirely valid way of implementing it would be to match it against the leaf nodes only, so the outer `description` in your example would not match, because it contains keys rather than input, while the inner `description` would match. Also, if the surrounding tree-ish structure is of the same depth for all entries, `--simplify` would furthermore remove the `key` from the structure as well, leaving just its parent.

pavelstudeny:
Ah, I see. OK for both --ignore-keys and --simplify then, as long as the simplify output is predictable.

eemeli:
Closing in favour of #178.

PR #177, opened 2017-06-30
Set parser's strictNumberSign option

nkovacs:
Add tests for parser fixes.

This needs https://github.com/messageformat/parser/pull/3


lib/compiler.js
```diff
@@ -154,6 +154,11 @@ Compiler.prototype.compile = function(src, lc, plurals) {
   if (typeof src != 'object') {
     this.lc = lc;
     var pc = plurals[lc] || { cardinal: [], ordinal: [] };
+    if (this.mf.strictNumberSign) {
+      pc.strictNumberSign = true;
+    } else {
+      pc.strictNumberSign = false;
+    }
```
eemeli: This is a bit verbose; should use something like `pc.strictNumberSign = !!this.mf.strictNumberSign` instead.

PR #181, opened 2017-11-18
Don't require `other` in plurals when plural functions have been set manually

eemeli:
According to the spec, all `plural`, `select`, and `selectordinal` instances should include an `other` case as a fallback. If, however, messageformat has been initialised with a custom pluralisation function (i.e. with `new MessageFormat({ [lc]: function })`), that function (and therefore the string being parsed) may use a different set of pluralisation categories than the standard zero/one/two/few/many/other. Therefore in that case we should trust that the user knows what they're doing, and not complain about the missing key.

This is relevant to my interests as I'm working on [gettext-to-messageformat](/eemeli/gettext-to-messageformat), and gettext uses language-specific integers for pluralisation categories, along with  an inline-provided pluralisation function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
   - "node"
-  - "iojs"
-  - "0.12"
-  - "0.10"
+  - "6"
+  - "4"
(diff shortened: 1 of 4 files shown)


eemeli:
I rebased this on the #182 branch to get the tests to pass; that should get merged before this.

Issue #187, opened 2018-03-14
Agile Translation workflows and MessageFormat


jamesbreen5:
Hey, 
I have hit an issue with using MessageFormat as part of a regular translation workflow, and I am looking for some advice - I am assuming my situation is not unique here.
Essentially, what the translation workflow does is inject English strings where no translation exists yet.
This causes all Asian languages to throw up errors with MessageFormat.compile, as they now contain the English one{..} category.
Is there anyway to ignore these errors, as the project essentially stops building until the translations come back in for the Asian languages and replace the English fallback.

SlexAxton:
If you handling doing the fallback, you’ll want to render fallback strings
in the english locale, not the one with missing strings.

It’s been a while though, do we do this fallback somewhere on your behalf?

eemeli:
We don't (yet...) provide fallback between locales, as the compiler output does not include an accessor method. Writing one is relatively simple, though; here's a sketch of something that might be appropriate here (note: haven't tested this specific code at all):

```js
// presuming that you're using messageformat/loader or precompiling your strings
import en from 'messages/en'
import ja from 'messages/ja'

const data = { en, ja }

class Messages {
  availableLocales = Object.keys(data)
  defaultLocale = 'en'
  locale = 'ja'
  has(key, lc) {
    const d = data[lc || this.locale]
    return d && typeof d[key] === 'function'
  }
  get(key, props) {
    if (this.has(key)) {
      return data[this.locale][key](props)
    } else if (this.locale !== this.defaultLocale && this.has(key, this.defaultLocale)) {
      return data[this.defaultLocale][key](props)
    } else {
      return key
    }
  }
}

export default new Messages
```

jamesbreen5:
Thanks, I can handle it on my end so, in how we fall back

PR #189, opened 2018-03-21
Add runtime accessor class

eemeli:
Re: #187, possibly some other issues as well.

This adds `lib/get.js`, a runtime accessor class for compiled MessageFormat objects. It's a new API, so the method and other names can and should be changed if we think of something else that's better. Here's how it would work:

First, presuming that we have the messages in a JSON file `messages.json`:
```json
{
  "en": {
    "a": "A {TYPE} example.",
    "b": "This has {COUNT, plural, one{one user} other{# users}}.",
    "c": {
      "d": "We have {P, number, percent} code coverage."
    }
  },
  "fi": {
    "b": "Tällä on {COUNT, plural, one{yksi käyttäjä} other{# käyttäjää}}.",
    "e": "Minä puhun vain suomea."
  }
}
```

And that we've compiled that into JavaScript:
```
messageformat -l en,fi messages.json > messageData.js
```

Then we can use the new getter and this API:
```js
var MessageFormatGet = require('messageformat/lib/get');
var msgData = require('./messageData');
var messages = new MessageFormatGet(msgData, 'en');

messages.hasMessage('a')                // true
messages.hasObject('c')                 // true
messages.get('b', { COUNT: 3 })         // 'This has 3 users.'
messages.get(['c', 'd'], { P: 0.314 })  // 'We have 31% code coverage.'

messages.get('e')                       // 'e'
messages.setFallback('en', ['foo', 'fi'])
messages.get('e')                       // 'Minä puhun vain suomea.'

messages.locale = 'fi'
messages.hasMessage('a')                // false
messages.hasMessage('a', 'en')          // true
messages.hasMessage('a', null, true)    // true
messages.hasObject('c')                 // false
messages.get('b', { COUNT: 3 })         // 'Tällä on 3 käyttäjää.'
messages.get('c').d({ P: 0.628 })       // 'We have 63% code coverage.'
```

Based on how I've used MessageFormat myself and seen it used by others, I believe that the above would provide a more developer-friendly interface, as well as being less likely to throw errors.

The code already has JSDoc comments, but here's an overview of the class's members:

```typescript
constructor (locales: object, defaultLocale: string | null | undefined)

availableLocales: string[]  // readonly
defaultLocale: string | null
locale: string

addLocale (lc: string, data: object)
getFallback (lc: string | undefined): string[]
setFallback (lc: string, fallback: string[] | null)

get (key: string | string[], props: object | undefined, lc: string | undefined): string | object
hasMessage (key: string | string[], lc: string | undefined, fallback: boolean | undefined): boolean
hasObject (key: string | string[], lc: string | undefined, fallback: boolean | undefined): boolean
```

For getters, `this.locale` is used if `lc` is null or undefined. `key` can be a string for top-level objects, or a keypath for hierarchical objects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ CHK=${GREEN} ✓${STOP}
 ERR=${RED} ✖${STOP}
 
 BIN=./node_modules/.bin
-SRC=lib/index.js lib/compiler.js lib/formatters/* lib/runtime.js
+SRC=lib/index.js lib/compiler.js lib/formatters/* lib/get.js lib/runtime.js
 
 .PHONY: all test test-browser doc release clean
 
--- a/lib/get.js
+++ b/lib/get.js
@@ -49,8 +49,14 @@
  * messages.get('c').d({ P: 0.628 })       // 'We have 63% code coverage.'
  */
 function Get(locales, defaultLocale) {
-  this._data = locales || {};
-  this._fallback = {}
+  this._data = {};
+  this._fallback = {};
+  Object.keys(locales).forEach(function(lc) {
+    if (lc !== 'toString') {
+      this._data[lc] = locales[lc];
+      if (typeof defaultLocale === 'undefined') defaultLocale = lc;
+    }
+  }, this);
 
   /**
    * List of available locales
@@ -65,7 +71,7 @@ function Get(locales, defaultLocale) {
    * Current locale
    * @member {string}
    */
-  this.locale = typeof defaultLocale !== 'undefined' ? defaultLocale : this.availableLocales[0];
+  this.locale = defaultLocale;
 
   /**
    * Default fallback locale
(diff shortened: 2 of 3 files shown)


SlexAxton:
I didn’t initially implement this because I assumed people would want to integrate their tooling in their own way with compiled messages and stuff, but:

a) we provide a lot more of that tooling now, especially with the cli stuff

b) this is super small and something everyone would have to end up implementing.

I assume adding this will be dead code for a lot of people, but I think it’s small enough to warrant the feature.

:+1:

eemeli:
I'll add some tests, and keep this open for at least a few days in case the API ends up warranting a bit of bikeshedding. I'm not sure I'm happy yet with the class name `Get` in particular.

Re: dead code, at least via `require()` and `import` this only gets included if it's implicitly asked for. And in any case it gzips to something like half a kilobyte, which is probably good enough. :)

eemeli:
Merged this to get it available in a `messageformat@next` release, but the API is still liable to change.

eemeli:
Noting some API changes here from the last week:
- The accessor class is now called `Messages`, and it's been moved from `lib/` to the root so that it'll be available via npm as `messageformat/messages`.
- `addMessages` replaces `addLocale`, to allow for partial updates of a locale.
- Setting `locale` and `defaultLocale` will match the given string to partial match of the available locales; they use the new method `resolveLocale` to do that.

PR #194, opened 2018-04-10
Add TypeScript definition file

MrDesjardins:
This PR contains a simple definition file that allows TypeScript developer to use the library without having to shim a custom definition file in their project. The definition file is not complete at the moment but sufficient for basic scenarios.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,15 @@
+declare module "messageformat" {
+    class MessageFormat {
+        constructor(message: { [pluralFuncs: string]: Function });
+        constructor(message: string[]);
+        constructor(message: string);
+        addFormatters: (format: {}) => MessageFormat;
+        disablePluralKeyChecks: () => MessageFormat;
+        setBiDiSupport: (enable: boolean) => MessageFormat;
+        setStrictNumberSign: (enable: boolean) => MessageFormat;
+        compile: (messages: string, locale?: string) => Msg;
+    }
+
+    type Msg = (params: {}) => string;
+    export = MessageFormat;
+}
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,15 +1,21 @@
 declare module "messageformat" {
+    type Msg = (params: {}) => string;
+    type Formatter = (val: any, lc: string, arg?: string) => string;
+    type SrcMessage = string | SrcObject;
+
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -0,0 +1,15 @@
+declare module "messageformat" {
+    class MessageFormat {
+        constructor(message: { [pluralFuncs: string]: Function });
+        constructor(message: string[]);
+        constructor(message: string);
+        addFormatters: (format: {}) => MessageFormat;
+        disablePluralKeyChecks: () => MessageFormat;
+        setBiDiSupport: (enable: boolean) => MessageFormat;
+        setStrictNumberSign: (enable: boolean) => MessageFormat;
+        compile: (messages: string, locale?: string) => Msg;
```
eemeli: This is missing the case of `messages` being an object, which is pretty important. Maybe something like this?
```ts
type SrcMessage = string | SrcObject;

interface SrcObject {
    [x: string]: SrcMessage;
}

compile: (messages: SrcMessage, locale?: string) => Msg;
```

index.d.ts
```diff
@@ -0,0 +1,15 @@
+declare module "messageformat" {
+    class MessageFormat {
+        constructor(message: { [pluralFuncs: string]: Function });
+        constructor(message: string[]);
+        constructor(message: string);
```
eemeli: `message` is in fact an optional parameter, that should be covered as well.

index.d.ts
```diff
@@ -0,0 +1,15 @@
...
+        addFormatters: (format: {}) => MessageFormat;
```
eemeli: I think this could be:
```typescript
type Formatter = (val: any, lc: string, arg?: string) => string;
addFormatters: (format: { [name: string]: Formatter }) => MessageFormat;
```

MrDesjardins:
Hello @eemeli I updated the definition file. 

So far, it covers only the messageformat.js file and not other files. I think it's good enough to start with TypeScript and we can improve it or other people can contribute. This is from the perspective that I wanted to consume your library in TypeScript and I had to create this definition file on my own. I believe it can remove some entry barrier by providing the TypeScript definition file even if not 100% complete.

eemeli:
Excellent, thank you!

PR #200, opened 2018-07-20
es6 style imports for typescript

mbarz:
With this proposal it should be possible to import this package in typescript projects using require or es6 module style

``` typescript
// old way with require
import MessageFormat = require('messageformat');

// new way with import from syntax
import * as MessageFormat from 'messageformat';

// cleanest way, that needed adjustment in messageformat.js
import { MessageFormat } from 'messageformat';
```

The last version needed a change in the javascript file and was not possible only changing the types definition. Therefore I nested MessageFormat in itself

```javascript
// lib/messageformat.js
MessageFormat.MessageFormat = MessageFormat;
module.exports = MessageFormat;
```

### DIFF
--- a/lib/messageformat.js
+++ b/lib/messageformat.js
@@ -363,5 +363,4 @@ MessageFormat.prototype.compile = function(messages, locale) {
   return result;
 }
 
-MessageFormat.MessageFormat = MessageFormat;
 module.exports = MessageFormat;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-declare module "messageformat" {
+declare namespace messageformat {
     type Msg = (params: {}) => string;
     type Formatter = (val: any, lc: string, arg?: string) => string;
     type SrcMessage = string | SrcObject;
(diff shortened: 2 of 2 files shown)


lib/messageformat.js
```diff
@@ -363,5 +363,5 @@ MessageFormat.prototype.compile = function(messages, locale) {
   return result;
 }
 
-
+MessageFormat.MessageFormat = MessageFormat;
```
eemeli: This looks pretty weird to me. Is this really a common pattern for TypeScript compatibility with CommonJS exports? Would a TS user really expect for `import { MessageFormat } from 'messageformat'` to "just work", or would it need to be documented?

mbarz: This is more about es6 modules than Typescript. With es6 imports, I expect to import a module containing classes and functions and not a class or a function directly (like the fs module in node).

In Angular (typescript) every import looks like `import { MyClass } from './myClass.ts'`

When doing this consequently with require one would have to do
```javascript
const MessageFormat = require('messageformat').MessageFormat;
```
as equivalent to
```typescript
import { MessageFormat } from 'messageformat';
```

To get this working and also keep it working with the simple require, I did this "pretty weird" hack... And because it also looked not so beautiful to me, I did this as a second commit, that can be rejected ;)

For my needs it would be enough to have the first commit, that enables me to do

```typescript
import * as MessageFormat from 'messageformat';
```
because I can not use require in my angular project

eemeli:
Then I think it'd probably be best to (for now) leave out the nested export. As is, the packaging really is just CommonJS, and expanding that to explicitly support TS and/or ES6 is a somewhat bigger change. If you could update the PR with that change, I'd be happy to merge it in and push out a new patch release.

Updating the codebase is starting to be a bit of a hassle due to its ES5-ness; it might well make sense to refactor it as ES6 classes and imports, and add a Babel transpile stage for continued npm/CommonJS support. As a part of that we could/should revisit the packaging for various environments.

PR #202, opened 2018-07-31
A Formater to translate with objects

YunnXairou:
I have created this formatter for my work, but I think some it will be useful for some peoples. With this you can use objects in your translation.
- to access **object.key1** you just need to use `{object, keys, key1}`
It would be great if it becomes an official formatter. Feel free to rename it as you please

Example
======

```
var mf = new MessageFormat(['en', 'fi']);

mf.compile('The key name of the object people is {people, keys, name}')({ people: { name: 'Bastian'}})
// 'The key name of the object people is Bastian'

mf.compile('{game, keys, player.name} is level {game, keys, player.level}')({ game: { player: {level: 42, name: Kirby}}})
// 'Kirby is level 42'
```

### DIFF
--- a/lib/formatters/index.js
+++ b/lib/formatters/index.js
@@ -32,5 +32,6 @@ module.exports = {
   date: require('./date'),
   duration: require('./duration'),
   number: require('./number'),
-  time: require('./time')
+  time: require('./time'),
+  keys: require('./keys')
 }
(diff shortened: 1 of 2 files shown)


eemeli:
While your formatter is fine, I'd suggest releasing it as a separate package on npm. The default formatters in `messageformat` are a subset of the formatters that [the spec](http://icu-project.org/apiref/icu4j/com/ibm/icu/text/MessageFormat.html) defines; extending built-in support for additional formatters would be rather questionable.

PR #204, opened 2018-08-04
Allow MessageFormat content in function parameters

### LINKED ISSUE #201: Nesting not supported with custom fortmatters
The Format Guide states this:

> All types of messageformat statements may be nested within each other, to unlimited depth

However, it doesn't work to nest anything within a custom formatter:

```js
const chalk = require('chalk');
const MessageFormat = require('messageformat');
const mf = new MessageFormat('en-GB');
mf.addFormatters({
  error: function(v, lc, p) { return chalk.red(p); }
});
const messages = mf.compile({
  describe: 'There were {numErrors, error, {# errors}}'
});

messages.describe({ numErrors: 5 })        // 'There were 5 errors', where '5 errors' prints as red
```

You could also imagine nesting a plural format in there, to get `'no error'`, `'one error'`, `'# errors'`, and so on, printed in red.

eemeli:
This would fix #201, but requires messageformat/parser#8 before it'll work -- see there for more details.

For testing, use `npm link` from the parser's `function-args` branch and then `npm link messageformat-parser` here to get the appropriate version.

### DIFF
--- a/docs/lib_compiler.js.html
+++ b/docs/lib_compiler.js.html
@@ -208,8 +208,12 @@ Compiler.prototype.token = function(token, plural) {
       }
       if (!this.mf.fmt[token.key]) throw new Error('Formatting function ' + JSON.stringify(token.key) + ' not found!');
       args.push(JSON.stringify(this.lc));
-      if (typeof token.param === 'string') {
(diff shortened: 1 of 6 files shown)


eemeli:
I updated the parser to v4, and rebased the branch on the master.

PR #206, opened 2018-08-17
Add configuration option to specify the input file's encoding to cli/messageformat.js

skress:
This PR adds a configuration option to specify the encoding that is used to read the input file. The default for that option remains 'latin1'.

For reference: since Java 9 properties files are loaded in UTF-8 encoding, see https://docs.oracle.com/javase/9/intl/internationalization-enhancements-jdk-9.htm for details.

### DIFF
--- a/cli/messageformat.js
+++ b/cli/messageformat.js
@@ -12,6 +12,7 @@ const glob = require('glob');
 const MessageFormat = require('messageformat');
 const nopt = require('nopt');
 const path = require('path');
+const uv = require('uv');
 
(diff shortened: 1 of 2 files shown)


eemeli:
To answer my own question, yes, [`uv`](https://www.npmjs.com/package/uv) looks like the right tool for this.

@skress, would you be able to use that to detect the encoding, replacing the config option?

skress:
Good idea to automatically detect the encoding and thanks for the pointer to uv. I've changed the PR accordingly.

PR #228, opened 2019-03-06
Fix dependencies and two broken tests

toolness:
Hello!  I was trying to get this to run on my system and noticed when I ran `npm install` (as well as `npm install --no-save`, which I _think_ made it install the exact dependencies in `package-lock.json`, but I could be wrong) it complained about a high severity vulnerability involving [prototype pollution in handlebars](https://nodesecurity.io/advisories/755) (I'm not sure if that link will actually work, it's currently timing out for me).

Anyways, it suggested I run `npm audit fix` and I did, and the vulnerabilities were no more!

So 36c4f47 addresses that.

However, aside from that--and regardless of whether 36c4f47 is applied--I also noticed that `npm test` fails on my system, specifically on two formatter tests that test date output.  It seems that when displaying dates, either MessageFormat or Node (I'm using 10.15.0 on Windows) is converting them to the local timezone.  I'm in GMT-5h, which meant that the dates were being displayed as the day before they were expected to (since the time portion for the dates was midnight).  I "fixed" this in 4a8ca30 by using [`Date.prototype.getTimezoneOffset()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset) to compensate for the difference in time, so that the times displayed are correct regardless of the system's local timezone.

That said, I'm not sure if this is the right fix/workaround.

Since both of the commits in this PR address entirely different things, I can split them into two separate PRs if you want, or I can discard one or both if you think my fix wasn't the right solution.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,56 +25,58 @@
       }
     },
     "@lerna/add": {
-      "version": "3.6.0",
-      "resolved": "https://registry.npmjs.org/@lerna/add/-/add-3.6.0.tgz",
-      "integrity": "sha512-aFVekkHMno3hj1Vg3EiIpAwrZ4g34i8z4KrCx7ATY6BRuxVT4Nt/Nk3l2k6gEOq3tWUDtUctLHxIAo14FI8sng==",
+      "version": "3.13.1",
+      "resolved": "https://registry.npmjs.org/@lerna/add/-/add-3.13.1.tgz",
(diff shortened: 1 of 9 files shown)


eemeli:
@toolness Would you be able to rebase your test fixes onto the current master, or would you be okay with someone else doing do? The `package-lock.json` issues should now be resolved as a part of #231.

toolness:
Ack, sorry I dropped the ball on this!  Happy to have someone else do it, but if no one else has the time I can give rebasing a shot...

eemeli:
Closing this, as I've ended up fixing the affected tests slightly differently in 8df5f78.

Issue #229, opened 2019-03-12
Compact Decimal Format


snewcomer:
👋 I worked on a solution a few months ago to add short number formatting to a library similar to this and have this working in some applications.  

https://github.com/snewcomer/cldr-compact-number

I'm not familiar with the internals of this library so I wanted to float the idea by you if you thought it might be a good addition.

Convert numbers like -
- `101234` to `101,1` mil in Espanol
- `101234` to `10.1万` in Japanese

https://github.com/ember-intl/intl-messageformat/pull/1

SlexAxton:
Definitely a good idea to mix these libraries, but don’t know if we should include them together by default.

We afford for adding custom short formatters in user-land. Perhaps we can document/ship a configuration that ties your lib to a short formatter in ours?

eemeli:
With the monorepo structure, it might make sense to split out the formatters into their own package, and to by default only import the ones that are actually part of the [spec](http://icu-project.org/apiref/icu4j/com/ibm/icu/text/MessageFormat.html). Then `compact` could be implemented as a custom argStyle of `number`.

snewcomer:
^^ I like that idea.  Lmk if/when a repo for `compact` is created and I can start working on it.  Would be happy to help porting over the other formatters as well.

eemeli:
@snewcomer This repo now includes `packages/formatters/`, which is the source for the soon-to-be-published `messageformat-formatters` package. Atm it just contains the default formatters; once `compact` or something like it is also included, we can work on the API for using them.

snewcomer:
@eemeli Perfect!  Yeah I've been tracking the work that has gone in! 👍  Hopefully within the next month I'll be able to work on this.

snewcomer:
@eemeli I'm not sure of an easy path to integrating compact number formatting.  It has a dependency on a large data set to be put into memory in order to translate a number to its compact equivalent.  What we do on a few ember projects is at build time generate a file based on the user's supported languages and then hydrate the locale data at runtime.

My initial reaction is this may be too large of a task/outside the bounds to implement for this library.  Let me know what you think.

https://github.com/unicode-cldr/cldr-numbers-full

eemeli:
@snewcomer Actually that sounds rather close to what messageformat is already doing when it compiles stringifiable output. That output includes plural category functions for just the included plurals.

If you were to put together a draft PR to achieve what you want without considering the efficiency concerns, I'd be happy to work with you at making it suitable for release.

[...]

PR #231, opened 2019-04-06
Refactor messageformat package as ES6, split formatters from it

eemeli:
This adds a Babel build stage to the `messageformat` package, allowing for the source to be written in modern JS. The web export is also now built with Webpack rather than Browserify + UglifyJS, and uses a different target config than the Node build (for now, the Browserslist defaults).

Simultaneously, formatters are getting split into their own package (Ping @snewcomer, re: #229).

The `reserved-words` dependency is dropped, as it was adding 15k to the minified size of the package built for web (by dragging in `assert` as its dependency).


snewcomer:
@eemeli This looks great!  Just starting to look at adding a short number formatter.  Curious - what ensures the short-number formatter doesn't get included in the user's build?  Is it the compile time work that only includes the necessary compiled functions?

eemeli:
At the moment the formatters' `index.js` does export all of the formatters, but that's not a necessity going foward. We may well include others like your short-number format in the library without importing them by default, making them available instead via a path like `messageformat-formatters/lib/compact-number`.

PR #242, opened 2019-06-30
Add option for array output

### LINKED ISSUE #241: Problem with Component Interpolation and V18n
We have an Issue concerning the message compiler. 
When trying to use v18n component interpolation with VueJs this does not work correctly due to the compile function which is returning a single string.
the function returns an `[object Object]` which shows us that r has elements of `VNodes` of Vue.
We were able to  fix the problem by using the following line in the message compiler, which returns an array which we can then use.
`return 'function(d) { return [' + (r.join(', ') || '""') + ']; }';`
Is it possible to add a configuration which allows us to determine whether we get a string or an array back by the function?

eemeli:
Fixes #241 

This adds a new method `MessageFormat#setArrayOutput()`, which makes compiled functions return arrays instead of strings, allowing for variable values to not be stringified. Some examples:

```js
const mf = new MessageFormat('en').setArrayOutput()

mf.compile('simple message')()
// ['simple message']

const obj = { foo: 'bar' }
mf.compile('objectively, {obj} is an object')({ obj })
// ['objectively, ', obj, ' is an object']

mf.compile('select {num, plural, one{one} other{#}}')({ num: 42 })
// ['select ', [42]]

### DIFF
--- a/packages/messageformat/src/compiler.js
+++ b/packages/messageformat/src/compiler.js
@@ -37,7 +37,7 @@ export default class Compiler {
       const pc = plurals[lc] || { cardinal: [], ordinal: [] };
       pc.strict = !!this.mf.options.strictNumberSign;
       const r = parse(src, pc).map(token => this.token(token));
-      return `function(d) { return ${r.join(' + ') || '""'}; }`;
+      return `function(d) { return ${this.concatenate(r, true)}; }`;
     } else {
       const result = {};
       for (var key in src) {
@@ -54,11 +54,19 @@ export default class Compiler {
     const r = token.cases.map(({ key, tokens }) => {
       if (key === 'other') needOther = false;
       const s = tokens.map(tok => this.token(tok, plural));
-      return propname(key) + ': ' + (s.join(' + ') || '""');
+      return `${propname(key)}: ${this.concatenate(s, false)}`;
     });
     if (needOther)
       throw new Error("No 'other' form found in " + JSON.stringify(token));
     return `{ ${r.join(', ')} }`;
+  }
+
+  /** @private */
+  concatenate(tokens, root) {
+    const asValues = this.mf.options.returnType === 'values';
+    return asValues && (root || tokens.length > 1)
+      ? '[' + tokens.join(', ') + ']'
+      : tokens.join(' + ') || '""';
   }
 
   /** @private */
--- a/packages/messageformat/src/messageformat.js
+++ b/packages/messageformat/src/messageformat.js
@@ -67,6 +67,9 @@ export default class MessageFormat {
    * @classdesc MessageFormat-to-JavaScript compiler
    * @param {string|string[]|Object} [locale] - The locale(s) to use
    * @param {Object} [options] - Compiler options
+   * @param {('string'|'values')} [options.returnType='string'] - Return type of
+   *   compiled functions; either a concatenated string or an array (possibly
+   *   hierarchical) of values
    * @param {boolean} [options.biDiSupport=false] - Add Unicode control
    *   characters to all input parts to preserve the integrity of the output
    *   when mixing LTR and RTL text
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by eemeli** [comment]
@FSchubert92 Would you be able to verify that this would solve your use case?

**Comment by rxaviers** [comment]
> ```js
> mf.compile('select {num, plural, one{one} other{#}}')({ num: 42 })
> // ['select ', [42]]
> ```
Why `['select ', [42]]` instead of `['select ', 42]`? Note `[42]` vs `42`.

Also, it could be useful to distinguish literal parts of the message vs variables... One approach is using an array of parts such as the one defined in Ecma-402 ([example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/formatToParts#Description)). For example:

```js
mf.compile('select {num, plural, one{one} other{#}}')({ num: 42 })
// [{ type: "literal", vaue: "select " }, { type: "variable", value: 42 }]
```

rxaviers:
Another comment: `MessageFormat#setArrayOutput()` modifies the behavior of the `compiler` returned function. Would it make sense to have another method that sort of function, e.g., `partsCompiler`?

rxaviers:
EDIT: *for* that...

eemeli:
Nested arrays: The runtime code is significantly simpler if the array isn't flattened for select cases. Now, admittedly for the given example the compiler could (and possibly should) skip the [wrapper], but it would still need to apply it for plural or select cases with a compound value. Also, as far as I can tell at least Vue and React don't really mind traversing a nested array.

`formatToParts`: Sure, that's another possibility, but is there an actual use case that should be considered for that? As is, at least React and probably other frameworks can use an array of values as is, but they couldn't use an array of `{ type, value }` quite as directly.

`compileParts`: While working on this, I started to realise that the current ability to mutate and persist a MessageFormat instance across configuration changes is a bit clumsy. So I'm drafting another PR that'll enable using an options object as the second arg for the constructor, with an intent of later deprecating and eventually removing the current set/disable methods. This change is also meant to more clearly communicate that a MessageFormat instance is relatively light to construct, and that it'd be fine to build more than one if really needing to use different sets of options at the same time.

Btw, just to be clear, I'm happy to continue this conversation and even happier if the shape of this PR improves as a result. My starting point here is to find a decent way to allow variables to pass through without stringification, but if we get to solve other problems along the way, that's even better.

FSchubert92:
@eemeli Yes, this option should fit our use case!

erickwilder:
My 0.02 cents:
> This change is also meant to more clearly communicate that a MessageFormat instance is relatively light to construct, and that it'd be fine to build more than one if really needing to use different sets of options at the same time.

I would say this could be easily handled as a further iteration and given the limited context of this change (allowing vue-i18n interpolation with a custom formatter) this PR is definitely good as it is _for now_

FSchubert92:
I just wanted to quickly ask whether there are some news on this issue?

eemeli:
This looks likely to become a part of the major release, of which I could probably package a beta/RC release within a week or so. Mostly because it'd be a bit silly to publish a new method just before removing such methods.

I'm still considering the exact API, to allow for the parts variant suggested by @rxaviers as a later addition without any breaking changes. Maybe something like this?
```js
const mf = new MessageFormat('en', { returnType: 'values' })
```
Alternative naming suggestions are rather welcome. The default value of the option would almost certainly be `'string'`, and `'values'` is the same as what this PR currently does, and later we might add `'parts'`. Not sure if the shape of the output ought to be reflected here as well with something like `'flat-values'`, as `[].flat()` is surprisingly [widely available](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) already, though not by IE or Edge.

eemeli:
Rebased on current `master` and refactored API to match that suggested in previous comment.

oppianmatt:
thanks for adding this, I see it's in the beta branch. seems the typings for typescript was missed with regards to this option

PR #244, opened 2019-07-17
Drop mutating methods + other breaking changes

eemeli:
This will require a major version bump. The following methods are dropped:
- `addFormatters()`
- `disablePluralKeyChecks()`
- `setBiDiSupport()`
- `setIntlSupport()`
- `setStrictNumberSign()`

Furthermore, locale-specific plural key checks are now always enforced when using default plural functions and Intl support is always assumed. For the other removed methods, corresponding options may be given to the constructor.

Quite a few internals get refactored as well to simplify/remove formatter and runtime caching during compilation, along which `MessageFormat.formatters`, `MessageFormat#fmt`, and  `MessageFormat#runtime` also get dropped. These had public visibility but were not explicitly documented as public.

I'll go ahead and publish a 2.3 release from the current master, to capture the preceding latest changes.


SlexAxton:
A little bit interesting in why we are moving `addFormatters` away from the setting and into the second arg, but I assume it's so you can do different formatters per locale or something? I'm fine with it either way.

:+1: anything else we want to put into a major version bump?

eemeli:
The main reason to also specify formatters in the constructor arg is API simplicity: this way the resulting instance has only one public method, `compile()`. Also, this allows for the instance to be immutable, which is just nice in general.

Regarding other changes to include in the version bump, there's at least #245 and #246. I'm not in any hurry on this, so happy to let this mellow for at least a week or two to see if anything else comes up.

PR #249, opened 2019-08-04
Always format # in plurals as number

### LINKED ISSUE #246: Should we support shorthand numbers inside of a simpleArg?
I don't know if this is cross-lib-compatible, but one thing I dislike is having to retype the plural variables in order to format them.

Any time I want to use `#` in a `pluralFormat` I also want to format that number, so `#` becomes useless.

Not suggesting we should yet, but wondering your thoughts, @eemeli on allowing:

```
// Old unformatted numbers
{numTotal, plural, one{# result} other{# results}}

// Old formatted numbers
{numTotal, plural, one{{numTotal, number} result} other{{numTotal, number} results}}

// Possible mashup
{numTotal, plural, one{{#, number} result} other{{#, number} results}}
```

eemeli:
Fixes #246 

BREAKING CHANGE: Previously, non-numeric values used as plural arguments could be referred to using `#` while keeping their original value. With this change, they now get converted to numbers and then formatted, probably coming out as `"NaN"` instead.

The number formatter relies on Intl.NumberFormat being available, and caches its instances.

### DIFF
--- a/packages/messageformat/src/compiler.js
+++ b/packages/messageformat/src/compiler.js
@@ -112,8 +112,12 @@ export default class Compiler {
       case 'octothorpe':
         if (!plural) return '"#"';
         fn = 'number';
-        args = [propname(plural.arg, 'd'), JSON.stringify(plural.arg)];
-        if (plural.offset) args.push(plural.offset);
+        args = [
+          JSON.stringify(this.lc),
+          propname(plural.arg, 'd'),
+          plural.offset || '0',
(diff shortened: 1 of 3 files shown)


SlexAxton:
Is it specified that `#` will be formatted with a default formatter? I assumed it would just be `toString`d

eemeli:
Yeah, the formatter is actually specified. See the bit I quoted [here](https://github.com/messageformat/messageformat/issues/246#issuecomment-514521319).

PR #250, opened 2019-08-04
Split messageformat-messages into its own package

eemeli:
Also refactor the source as an ES6 class, and use [jsdoc2md](https://www.npmjs.com/package/jsdoc-to-markdown) to generate its documentation.

The main rationale for this is to allow for `messageformat` itself to be used as a dev dependency even if the accessor class is used as well.


eemeli:
As a heads-up, I kinda went on from here and re-refactored this into `messageformat-runtime`, which also includes well, the runtime methods. That's a bit in-progress still, will add a PR for it later.

PR #254, opened 2019-08-28
Fixed typo when setting default locale, removed double export

bobvandevijver:
The default locale was not set correctly due to a missing underscore (this was also the case with the js version of this file). I've also removed the double export, as that resulted in a "Cannot access 'Messages' before initialization", which seems logical.

### DIFF
--- a/packages/messages/src/messages.js
+++ b/packages/messages/src/messages.js
@@ -8,8 +8,6 @@
  * @module messageformat-messages
  */
 
-module.exports = Messages;
-
 /**
  * @alias module:messageformat-messages
  *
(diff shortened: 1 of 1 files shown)


eemeli:
The double export is a valid bug; it was probably only working in testing due to the class getting transpiled into a function, which then got hoisted and hid the error. Thank you for noticing this!

The line `this.locale = defaultLocale` in the constructor was intentional, though. What that does is pass `defaultLocale` to the `locale` setter, which uses `resolveLocale` to clean up the value that actually gets assigned to `this._locale`. Admittedly it does look a bit misleading, esp. with the subsequent line then not using the `locale` getter.

bobvandevijver:
Hmm, not sure what happened during my tests a couple days back then (and I cannot seem to reproduce it anymore as well 🤷‍♂)... 
Anyways, I've updated the MR to now use the locale getter instead to make it a bit less confusion 😃

PR #257, opened 2019-09-01
Refactor API for v3

eemeli:
I started out trying to do this in smaller parts, but I couldn't get the browser tests to pass until I had most of it done. So, with apologies, this got kinda huge. The common themes here are that:
1. The API presented by our default export looks much more like `Intl` APIs.
2. The module output is now explicitly a string representation of an ES module which has runtime dependencies, allowing for tree-shaking to keep the final compiled output truly minimal. 

To see some of the changes, run `npm run docs:example` from the `refactor` branch to build `example/messages.mjs` (ES module) and `example/messages.js` (minified, then prettied).

Here's how the API looks with this set of changes:

```js
import MessageFormat from 'messageformat'
```

#### `new MessageFormat(locales, options)`
For custom plurals, `locales` should now be an array of named functions, rather than an object. Using the special value `'*'` will match all available locales. Also, custom locale functions may have a `module` property set to mark them as runtime-importable.

`options` has a new entry `currency` (default `'USD'`), used when formatting `{var, number, currency}`

#### `mf.compile(message): Function`
Now only compiles a single string message into a corresponding function. If called on an instance with multiple plurals, uses the first one.

#### `MessageFormat.supportedLocalesOf(locales): string[]`<br>`mf.resolvedOptions(): Object`
New methods, in the style of similarly-named Intl object methods.

```js
import compileModule from 'messageformat/compile-module'
```

#### `compileModule(messageformat, messages): string`

This is separated from the MessageFormat class in order to leave it out from the default browser-targeted build, as this doesn't really make any sense as a runtime function. It takes in a `messages` object, and outputs the string representation of an ES module, which when executed as JavaScript will export an object with a matching shape, consisting of functions. This module will `import` some of its dependencies from `messageformat-runtime`, a new package that merges `messageformat/src/runtime.js`, `messageformat-formatters`, and `messageformat-messages`.

As `messageformat-runtime` is a dependency of `messageformat`, in most cases it'll "just work". The runtime is published as both CommonJS (\*.js) & ES modules (\*.mjs), in order to allow tree-shaking to get rid of any unused parts. This means that builds with `webpack --mode development` may include up to 50Kb more cruft than `webpack --mode production`, as the former does not apply tree-shaking. Mostly this is due to the 200+ locale category functions from `make-plural`.


eemeli:
@SlexAxton I'm going to merge this later today or tomorrow in order to get the v3 release forward a bit.

Issue #255, opened 2019-08-28
Typescript typing for messageformat/messages


bobvandevijver:
I'm not sure how you want these (do you create them manually?), but I created these typing:

```
declare module 'messageformat/messages' {
    /**
     * @classdesc Accessor for compiled MessageFormat functions
     *
     * ```
     * import Messages from 'messageformat/messages'
     * ```
     *
     * @class
     * @param {object} locales A map of locale codes to their function objects
     * @param {string|null} [defaultLocale] If not defined, default and initial locale is the first entry of `locales`
     *
     * @example
     * var fs = require('fs');
     * var MessageFormat = require('messageformat');
     * var mf = new MessageFormat(['en', 'fi']);
     * var msgSet = {
     *   en: {
     *     a: 'A {TYPE} example.',
     *     b: 'This has {COUNT, plural, one{one user} other{# users}}.',
     *     c: {
     *       d: 'We have {P, number, percent} code coverage.'
     *     }
     *   },
     *   fi: {
     *     b: 'Tällä on {COUNT, plural, one{yksi käyttäjä} other{# käyttäjää}}.',
     *     e: 'Minä puhun vain suomea.'
     *   }
     * };
     * var cfStr = mf.compile(msgSet).toString('module.exports');
     * fs.writeFileSync('messages.js', cfStr);
     *
     * ...
     *
     * var Messages = require('messageformat/messages');
     * var msgData = require('./messages');
     * var messages = new Messages(msgData, 'en');
     *
     * messages.hasMessage('a')                // true
     * messages.hasObject('c')                 // true
     * messages.get('b', { COUNT: 3 })         // 'This has 3 users.'
     * messages.get(['c', 'd'], { P: 0.314 })  // 'We have 31% code coverage.'
     *
     * messages.get('e')                       // 'e'
     * messages.setFallback('en', ['foo', 'fi'])
     * messages.get('e')                       // 'Minä puhun vain suomea.'
     *
     * messages.locale = 'fi'
     * messages.hasMessage('a')                // false
     * messages.hasMessage('a', 'en')          // true
     * messages.hasMessage('a', null, true)    // true
     * messages.hasObject('c')                 // false
     * messages.get('b', { COUNT: 3 })         // 'Tällä on 3 käyttäjää.'
     * messages.get('c').d({ P: 0.628 })       // 'We have 63% code coverage.'
     */
    class Messages {

        /**
         * List of available locales
         * @readonly
         * @memberof Messages
[...]

PR #265, opened 2019-10-17
Added typings for messageformat-runtime/messages

bobvandevijver:
As requested in #255 by @eemeli , the typing I created as PR for the current master branch. 

If you expect them to have a different layout (maybe you want the comments stripped here?), let me know.

### DIFF
--- a/packages/runtime/messages.d.ts
+++ b/packages/runtime/messages.d.ts
@@ -1,0 +1,45 @@
+declare module 'messageformat-runtime/messages' {
+  class Messages {
+    public availableLocales: string[];
+
+    public locale: string | null;
+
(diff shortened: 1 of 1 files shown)


eemeli:
I think leaving out the comments might make sense, unless that controls whether they're used/visible by autocompletion tools? At the very least any duplication of TS types in the JSDoc comments should be dropped.

bobvandevijver:
They shouldn't influence the autocompletion I guess, although they might influence what text is shown there. You do not have the comment in https://github.com/messageformat/messageformat/blob/master/packages/messageformat/index.d.ts either, so I've now removed them completely.

PR #269, opened 2019-11-25
[WIP] feat: cldr compact format

snewcomer:
I'd hate to submit such an incomplete PR but I wanted to get the work started to ensure I see it through.  I also need some help to understand how we can include this as a new API.  Some questions I have on the outset...

1. What dependencies would we need to compile stringifiable output of just the included locales?   Do we need to create a separate build package?
2. What entry point could we hook in a compact format function?  Any specific files I can take a look at?

Thank you for the help!!

https://github.com/snewcomer/cldr-compact-number

### DIFF
--- a/packages/formatters/package.json
+++ b/packages/formatters/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "messageformat-formatters",
+  "version": "0.1.0",
+  "description": "Formatters for messageformat",
+  "keywords": [
+    "icu",
+    "messageformat",
+    "formatter"
+  ],
+  "author": "Eemeli Aro <eemeli@gmail.com>",
+  "license": "MIT",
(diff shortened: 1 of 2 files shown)


packages/formatters/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "url": "https://github.com/messageformat/messageformat.git",
+    "directory": "packages/formatters"
+  },
+  "main": "index.js",
+  "files": [
+    "index.js",
+    "lib/"
+  ],
+  "dependencies": {
+    "cldr-compact-number": "^0.4.0"
```
snewcomer: Screen Shot 2019-11-24 at 9 49 00 PM

Note I include en locale information by default, so it is usable out of the box.

PR #272, opened 2019-12-04
Add support for number patterns & skeletons

eemeli:
We claim to support everything that ICU MessageFormat supports, but that's not actually true. We've been missing `argStyleText` and the recently added [`argSkeletonText`](https://github.com/unicode-org/icu/blob/master/docs/userguide/format_parse/numbers/skeletons.md) support for date and number formatting. This PR adds that for numbers, so that the following messages will now work:

```
Your balance is {VAR, number, ¤#,##0.00;(¤#,##0.00)}
The sparrow flew {VAR, number, :: measure-unit/length-meter unit-width-full-name}.
```

Producing output like:
```
Your balance is ($3.27)
The sparrow flew 42 meters.
```

To enable that, I've added a new repo [messageformat/number-skeleton](https://github.com/messageformat/number-skeleton) with a parser & formatter-generator for pattern & skeleton strings. It's a separate repo because it's in TypeScript, unlike the rest of the codebase. It may also be useful for other libraries, as it's not really locked to `messageformat`.

The formatters are rather dependent on `Intl.NumberFormat`, and in some cases depend on options that are only supported in very recent Chromium releases. Even then, not everything that's possible with patterns or skeletons works in JavaScript.


eemeli:
Had to adjust tests a bit to get them to pass in all environments.

One major downside of this change is that it'd increase the gzipped minified size of the compiler from 16kB to 22kB, which is non-trivial. For build-time compilation that doesn't matter, of course, but it's still a rather common way of using the library. Maybe some sort of factory functionality ought to be added, a bit like I've done with [intl-pluralrules](https://github.com/eemeli/intl-pluralrules#factory)?

SlexAxton:
I would bias towards shipping it and pulling it out later, personally. Those two sizes seem reasonable for the functionality, especially since size-conscious folks should not be shipping the runtime.

PR #276, opened 2020-02-16
fix(contributing): use direct link to license agreement to messageformat

erickwilder:
The previous link [`https://js.foundation/CLA/`](https://js.foundation/CLA/) no longer provides information about CLA agreement/signing


eemeli:
Thank you! We should be soon migrating in any case to use the OpenJS Foundation CLA, but that's not available yet.

PR #291, opened 2020-09-13
Import date-skeleton & number-skeleton sources here

eemeli:
Previously, the date-skeleton and number-skeleton sources have been hosted under the [`messageformat/skeletons`](https://github.com/messageformat/skeletons) repository, as at the time they were the only TypeScript parts of the project.

Now that most of the source is TypeScript, it doesn't make sense to keep them separate, so let's merge them here.


eemeli:
Rebased on master & applied same rename as for other packages.

PR #297, opened 2020-11-16
Add Java .properties support to the Webpack loader

eemeli:
This will bring the loader to feature parity with the Rollup plugin, and allow for the [messageformat-properties-loader](https://www.npmjs.com/package/messageformat-properties-loader) package to be deprecated.


README.md
```diff
@@ -14,7 +14,6 @@ This monorepo consists of the following packages that make up our JS implementat
 - [@messageformat/convert](packages/convert/) - Converts other localization formats into MessageFormat
 - [@messageformat/date-skeleton](packages/date-skeleton) - Tools for working with [ICU DateFormat skeletons]
 - [@messageformat/loader](packages/webpack-loader/) - Webpack loader for JSON & YAML message files
```
cletusw, line 16:
```suggestion
- [@messageformat/loader](packages/webpack-loader/) - Webpack loader for JSON, YAML, & .properties message files
```

packages/webpack-loader/README.md
```diff
@@ -33,57 +31,53 @@ If you’re using Webpack 4, you must include `type: 'javascript/auto'` in the l
...
+messages['ordinal']({ N: 1 });
 // => 'The 1st message.'
 ```
 
 ## Options
 
-- [`locale`] The [CLDR language code] or codes to pass to [`messageformat`]. If using multiple locales at the same time, exact matches to a locale code in the data structure keys will select that locale within it (as in [`example/src/messages.json`](example/src/messages.json)). Defaults to `en`.
-- [`convert`] Use `@messageformat/convert` to convert non-MessageFormat syntax and plural objects into MessageFormat. Use an object value to configure. Defaults to `false`.
+- `convert` Use [`@messageformat/convert`][convert] to convert non-MessageFormat syntax and plural objects into MessageFormat. Use an object value to configure. Defaults to `false`.
+- `encoding` File encoding. Defaults to `'auto'`, which will auto-detect `'utf8'` and otherwise use `'latin1'`.
```
cletusw, line 67:
```suggestion
- `encoding` File encoding. Defaults to `'auto'`, which will attempt to detect `'utf8'`, falling back to `'latin1'`.
```

packages/webpack-loader/index.js
```diff
@@ -20,28 +22,33 @@ function localeFromResourcePath(resourcePath, locales) {
...
     locale = locale.split(',');
   if (Array.isArray(locale) && locale.length > 1) {
-    var relPath = path.relative(process.cwd(), this.resourcePath);
+    const relPath = relative(process.cwd(), this.resourcePath);
     locale = localeFromResourcePath(relPath, locale);
   }
-  const mfOpt = {};
-  if (options.biDiSupport) mfOpt.biDiSupport = true;
-  if (options.customFormatters || options.formatters) {
-    mfOpt.customFormatters = options.customFormatters || options.formatters;
```
cletusw, line 42: Does messageformat/core respect `formatters` now? Otherwise this will be a breaking change, albeit in an undocumented feature since we never listed `formatters` in the [README](https://github.com/messageformat/messageformat/tree/df713b1714837b92e2bf4c07d5323c21c78df278/packages/webpack-loader#options).

cletusw (approved):
Looks great!

packages/webpack-loader/index.js
eemeli: You're right, but it should be fine in this case. Due to the overall v3 update it seems like a good time to finally release this as a 1.0.0, and with #290 the package is even getting renamed to `@messageformat/loader`. So it's a good time for nominal breaking changes.

The remaining question is whether we need to explicitly mention the deprecation of an undocumented feature in the release notes, and I'd suggest that we're probably ok not doing so.

cletusw: Agreed, no doc should be fine especially with a major version bump. Kiitos!

PR #298, opened 2020-11-16
Add `rollup-plugin-messageformat` to README's list of packages

cletusw:
Note: I'm not sure if this is supposed to be `@messageformat/rollup-plugin` -- I did it this way to match the [top-level package.json](https://github.com/messageformat/messageformat/blob/df713b1714837b92e2bf4c07d5323c21c78df278/package.json#L41-L52)

Also definitely squash merge to remove all my intermediate commits -- I was using the online GitHub editor and couldn't find an amend button :)


eemeli (approved):
Thank you, good catch! The package is named without the `@messageformat/` org in order to match the [Rollup plugin conventions](https://rollupjs.org/guide/en/#conventions), which require `rollup-plugin-` as a prefix.

Issue #303, opened 2021-03-05
Failed to import make-plurals runtime


gamedevsam:
I ran into this issue today, simple variable substitution works fine, but when I add pluralization, I run into the following webpack error: 

```
> export * from 'make-plural/plurals';
| 
Module parse failed: 'import' and 'export' may appear only with 'sourceType: module' (6:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|  */
| /** @private */
> export function _nf(lc) {
|     // @ts-ignore
|     return _nf[lc] || (_nf[lc] = new Intl.NumberFormat(lc));

error Command failed with exit code 1.
```

Any ideas what could be causing this issue?

I'm using webpack with LWC (lightning-web-components) configs so the entire webpack config is a bit opaque, but I'll try to get that tomorrow to see if I can provide more info.

gamedevsam:
**UPDATE:** Removing `"type": "commonjs",` (line 24) of `@messageformat/runtime/package.json` fixed the error for me.

eemeli:
Sorry for taking so long to get to this. Could you test if the issue would also be resolved by adding a file `@messageformat/runtime/esm/package.json` with just the contents `{ "type": "module" }`?

PR #305, opened 2021-03-17
Remove type from package.json

gamedevsam:
This allows webpack and other bundlers to determine which module style to consume. Fixes https://github.com/messageformat/messageformat/issues/303


gamedevsam:
Closing in favor of https://github.com/messageformat/messageformat/pull/308

PR #308, opened 2021-03-20
fix(runtime): Add missing esm/package.json

### LINKED ISSUE #303: Failed to import make-plurals runtime

eemeli:
Fixes #303 
Closes #305

Ping @gamedevdan

In testing, I was able to duplicate the original reported issue with Webpack 5 and resolve it by adding the missing `esm/package.json` file that correctly identifies files within that folder as using ES rather than CommonJS modules.

Given that this should also enable `@messageformat/runtime/esm/...` content to be used in Node.js environments, it's much preferable to the alternative of removing the `"type"` from the module's root package.json file.

### DIFF
--- a/packages/runtime/esm/package.json
+++ b/packages/runtime/esm/package.json
@@ -1,0 +1,1 @@
+{ "type": "module" }


gamedevsam:
Testing your fix.

gamedevsam (approved):
This fixes it, ship it!

gamedevsam:
@eemeli can you make a new beta release that includes this change so I can get rid of my patch?

Thanks!

eemeli:
Yeah, I figure I'll do one around the end of the week, after I've merged the two currently-open PRs.

gamedevsam:
Thank you!

gamedevsam:
@eemeli any idea when a new release will go out? No pressure, just want to have an idea when I should check back in.

eemeli:
No strict plan, but no real reason to wait either. I'll see about getting a release out next weekend.

PR #314, opened 2021-04-07
Doc fixes

floratmin:
Fixed `rollup.config.js` and the conversion to string for usage in compile. The emitted file `messages.js` in the `@messageformat/runtime ` example is also broken.

### DIFF
--- a/docs/api/runtime.messages.md
+++ b/docs/api/runtime.messages.md
@@ -48,7 +48,7 @@ const msgSet = {
     e: 'Minä puhun vain suomea.'
   }
 };
-writeFileSync('messages.js', String(mf.compile(JSON.stringify(msgSet))));
+writeFileSync('messages.js', String(mf.compile(msgSet)));
 
 ```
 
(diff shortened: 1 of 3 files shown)


docs/api/runtime.messages.md
```diff
@@ -48,7 +48,7 @@ const msgSet = {
     e: 'Minä puhun vain suomea.'
   }
 };
-writeFileSync('messages.js', String(mf.compile(msgSet)));
+writeFileSync('messages.js', String(mf.compile(JSON.stringify(msgSet))));
```
eemeli, line 51: This is not the right fix, for two reasons:
1. These API docs are generated by api-extractor/api-documenter from source code comments, so this change needs to be applied to `packages/runtime/src/messages.ts`.
2. This is a remnant of the old API, where the `compile()` method accepted object input. That functionality has been moved to `@messageformat/core/compile-module`, which ought to be used here instead.

eemeli (changes requested):
Thank you for noticing these! The runtime messages example needs to be fixed a bit differently, though.

eemeli (approved):

eemeli:
Thank you!

PR #317, opened 2021-04-12
Update cli.md

arturmuller:
When I ran `npx messageformat messages/strings.json --outfile=messages/en.js` I get an error from npx saying `could not determine executable to run`.

Changing the command to use the `@messageformat/cli` package fixes the issue for me.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -52,7 +52,7 @@ Command-line options override configuration files.
 With `messages/strings.json`, compile it into an ES module using the default English locale:
 
 ```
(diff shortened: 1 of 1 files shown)


eemeli (approved):
This looks like a npm v7 change. Would be nice if there was a cleaner way, but we're probably stuck with this. :/

eemeli:
@arturmuller Did you try to run the `npx messageformat` command before installing `@messageformat/cli` locally? I'm having trouble replicating your experience any other way.

PR #319, opened 2021-04-28
Fix pathSep types

aeirola:
The `@messageformat/react` package `MessageProvider` prop `pathSep` types only allow `string` and `undefined` values, while the documentation states that `null` is also a valid value. This fixes the issue by adding the missing type.


eemeli (approved):

eemeli:
@aeirola This is pending on getting a CLA signature from you. Would it be possible for you to sign it? Just click the "Details" link next to the licence/cla check and follow the instructions.

aeirola:
Oh, yeah, absolutely. Sorry for not noticing that earlier. Thanks for bringing it up. Should be in order now.

PR #344, opened 2021-09-21
Use a string param in the select runtime function

Marr11317:
The select function should have a string type for the `value` parameter. Since it cannot be type checked, it doesn't throw any error, except when using the cli's output in ts files.


eemeli (approved):
Thank you for catching this!

PR #345, opened 2021-09-27
Support json5 in cli

Marr11317:
I would like to add tests or test it on my own, even for this relatively small change, but it seems that there is no infrastructure to do so. 
Is there a recommended way to test the pr?


Marr11317:
Do you think there's something missing before this can be merged?

eemeli:
The missing thing was a few minutes of time & effort on my part. Apologies for the delay.

Issue #350, opened 2021-12-14
Extend MessageFunction param typings


dpchamps:
[MessageFunction](https://github.com/messageformat/messageformat/blob/master/packages/core/src/messageformat.ts#L18) is typed like so:

```typescript
export type MessageFunction<ReturnType extends 'string' | 'values'> = (
  param?: Record<string, unknown>
) => ReturnType extends 'string' ? string : unknown[];
```

However, icu supports index lookup. For example, I could specify something like

```text
On {1,date,::dMMMM}, there was {2} on planet {0,number,integer}.
```

And this will produce valid, expected results:

```typescript
  const mf = new MessageFormat("*");
  const result = mf.compile([42, new Date("5/25/1977"), "a disturbance in the force"]);

  console.log(result) // On May 25, there was a disturbance in the force on planet 42.  
```

The type for the parameter should be `Record<string, unknown> | unknown[]`

eemeli:
Fixed by #351

PR #351, opened 2021-12-14
fix: update typing for MessageFunction param

dpchamps:
## Problem

Outlined https://github.com/messageformat/messageformat/issues/350

## Solution

Update type


eemeli:
Thank you, this looks like a valid and necessary type relaxation!

PR #353, opened 2022-01-05
Add check to detect test env when consumer is running tests

Johannesklint:
When the consumers of the library `messageformat/react` is writing tests for their component, which uses `useMessage` – a lot of `console.warn` is shown all over if there is no message provided.

Example:

```js
// component.jsx
function Component() {
  // We would get an console.warn here for both tests.
  // Since the second text is missing in the first test block and vice versa in the second test block
  const firstText = useMessage("first.text");
  const secondText = useMessage("second.text");
  return (
    <div>
      <p>{firstText}</p>
      <p>{secondText}</p>
    </div>
  );
}

// component.test.jsx
import { screen, render } from "@testing-library/react";
import { MessageProvider } from "@messageformat/react";

// just an example of how a consumers custom rendered could look like
function customRender(ui, options) {
  return render(
    <MessageProvider locale="sv" messages={options.messages}>
      {ui}
    </MessageProvider>
  );
}

test("first text is rendered", () => {
  customRender(<Component />, {
    message: {
      "first.text": "first text value"
    }
  });
  // here console.warns would be shown
  expect(screen.getByText("first text value")).toBeInTheDocument();
});

test("second text is rendered", () => {
  customRender(<Component />, {
    message: {
      "second.text": "second text value"
    }
  });
  // here console.warns would be shown
  expect(screen.getByText("second text value")).toBeInTheDocument();
});
```

With this small fix, no warnings would be shown for consumers tests

### DIFF
--- a/packages/react/src/message-provider.ts
+++ b/packages/react/src/message-provider.ts
@@ -65,7 +65,9 @@ function getOnError(
     throw new MessageError(path, code, asId);
   }
   function msgWarning(path: string[], code: ErrorCode) {
-    console.warn(errorMessages[code], path);
+    if (process.env.NODE_ENV !== "test") {
+      console.warn(errorMessages[code], path);
+    }
     return asId(path);
   }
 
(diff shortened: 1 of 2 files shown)


Johannesklint:
This solved with the `onError="silent"` prop: 
```js
<MessageProvider {...props} onError="silent">..</MessageProvider>

```

Issue #359, opened 2022-02-24
bug: incorrect locales used in in generated date functions


cdaringe:
# Problem

locale `es-MX` gets `en` formatted functions

# Demonstration

tiny runnable demo here: https://github.com/cdaringe/multi-locale-invalid-dates

it can be seen in just this code too:

```js
const MessageFormat = require("@messageformat/core");
const compileMessageModule = require("@messageformat/core/lib/compile-module");

const messagePacks = {
  "en-US": {
    utilsDate: "{date, date, ::EEEMMMd}",
    utilsToday: "today",
    utilsTomorrow: "tomorrow",
  },
  "es-MX": {
    utilsDate: "{date, date, ::MMMMd}",
    utilsToday: "hoy",
    utilsTomorrow: "mañana",
  },
};

const mf = new MessageFormat("*");
console.log(compileMessageModule(mf, messagePacks));
```

which emits:

```js
const date_en_MMMMd_meky2n = (function() { // should be es_mx
  var opt = {"month":"long","day":"numeric"}; // need to research if this is correct or not
  var dtf = new Intl.DateTimeFormat("en", opt); // should be "es-MX"
  return function(value) { return dtf.format(value); }
})();
export default {
  "en-US":   {/* snip */},
  "es-MX": {
    utilsDate: (d) => date_en_MMMMd_meky2n(d.date),
    utilsToday: () => "hoy",
    utilsTomorrow: () => "mañana"
  }
}
```

cdaringe:
looks to be that core wants `en` & `es`, not the locales 🤔

eemeli:
Confirmed. This mostly affects the `'*'` special-case locale code. It's not just the date functions, but the locale isn't detected at all from the object path like it should be.

This works like it should:
```js
const mf = new MessageFormat(['en-US', 'es-MX'])
console.log(compileModule(mf, { 'es-MX': { foo: '{x, plural, one{one} other{other}}' } }))
// import { es } from "@messageformat/runtime/lib/cardinals";
// import { plural } from "@messageformat/runtime";
// export default {
//   "es-MX": {
//     foo: (d) => plural(d.x, 0, es, { one: "one", other: "other" })
//   }
// }
```

But this sort of doesn't:
```js
const mf = new MessageFormat(['en-US', 'es'])
console.log(compileModule(mf, { 'es-MX': { foo: '{x, plural, one{one} other{other}}' } }))
// import { en } from "@messageformat/runtime/lib/cardinals";
[...]

PR #360, opened 2022-02-24
feat(core): Support locale variants in messages object

dpchamps:
Ref: https://github.com/messageformat/messageformat/issues/359

It was discovered that when compiling against locale variants (e.g. `en-US`, `es-MX` etc etc...) output is generated against expectations. 

For example

```javascript
const MessageFormat = require("@messageformat/core");
const compileMessageModule = require("@messageformat/core/lib/compile-module");

const messagePacks = {
  "en-US": {
    utilsDate: "{date, date, ::EEEMMMd}"
  },
  "es-MX": {
    utilsDate: "{date, date, ::EEEMMMd}"
  },
};

const mf = new MessageFormat("*");
const result = compileMessageModule(mf, messagePacks);
```

Where result is

```javascript
const date_en_EEEMMMd_tmb11z = (function() {
  var opt = {"weekday":"short","month":"short","day":"numeric"};
  var dtf = new Intl.DateTimeFormat("en", opt);
//------------------------------------^ this is problematic
  return function(value) { return dtf.format(value); }
})();

export default {
  "en-US": {
    utilsDate: (d) => date_en_EEEMMMd_tmb11z(d.date)
  },
  "es-MX": {
    utilsDate: (d) => date_en_EEEMMMd_tmb11z(d.date)
  }
}
```

The expectation is instead

```javascript
const date_en_EEEMMMd_tmb11z = (function() {
  var opt = {"weekday":"short","month":"short","day":"numeric"};
  var dtf = new Intl.DateTimeFormat("en", opt);
  return function(value) { return dtf.format(value); }
})();

const date_es_EEEMMMd_m7z124 = (function() {
  var opt = {"weekday":"short","month":"short","day":"numeric"};
  var dtf = new Intl.DateTimeFormat("es", opt);
  return function(value) { return dtf.format(value); }
})();

export default {
  "en-US": {
    utilsDate: (d) => date_en_EEEMMMd_tmb11z(d.date)
  },
  "es-MX": {
    utilsDate: (d) => date_es_EEEMMMd_m7z124(d.date)
  }
}
```

Variants are an expected part of locale strings, so it makes sense to support them inside of imports. This bug was caught in our codebase because we want to maintain parity with the icu variant spec.

The patch proposed here is (i think) pretty un-controversial: if a given plural key can be normalized, perform the lookup based on the normalized value.

### DIFF
--- a/packages/core/src/compile-module.test.ts
+++ b/packages/core/src/compile-module.test.ts
@@ -368,7 +368,7 @@ describe('compileModule()', function () {
       expect(m['es-MX']['key']()).toEqual('en-MX');
     });
 
-    it('date regression specification', async () => {
+    it('date regression specification', () => {
       const mf = new MessageFormat('*');
       const mp = {
         'en-US': {
--- a/packages/core/src/compile-module.test.ts
+++ b/packages/core/src/compile-module.test.ts
@@ -329,4 +329,63 @@ describe('compileModule()', function () {
       );
     });
   });
+
+  describe('locale variants', () => {
+    it('valid variants', async () => {
+      const mf = new MessageFormat("*");
+      const mp = {
+        "en-US": {
+          key: 'en-US'
+        },
+        "es-MX": {
+          key: "en-MX"
+        },
+        "es-ES": {
+          key: "en-ES"
+        }
+      }
+
+      const m = await getModule(mf, mp);
+
+      expect(m['en-US']['key']()).toEqual("en-US");
+      expect(m['es-MX']['key']()).toEqual("en-MX");
+      expect(m['es-ES']['key']()).toEqual("en-ES");
+    });
+
+    it('valid variants', async () => {
+      const mf = new MessageFormat("*");
+      const mp = {
+        "en-US": {
+          key: 'en-US'
+        },
+        "es-MX": {
+          key: "en-MX"
+        },
+        "es-ES": {
+          key: "en-ES"
+        }
+      }
+
+      const m = await getModule(mf, mp);
+      expect(m['en-US']['key']()).toEqual("en-US");
+      expect(m['es-MX']['key']()).toEqual("en-MX");
+      expect(m['es-ES']['key']()).toEqual("en-ES");
+    });
+
+    it('mixed variants', async () => {
+      const mf = new MessageFormat("*");
+      const mp = {
+        "en": {
+          key: 'en'
+        },
+        "es-MX": {
+          key: "en-MX"
+        },
+      }
+
+      const m = await getModule(mf, mp);
+      expect(m['en']['key']()).toEqual("en");
+      expect(m['es-MX']['key']()).toEqual("en-MX");
+    });
+  })
 });
--- a/packages/core/src/compile-module.test.ts
+++ b/packages/core/src/compile-module.test.ts
@@ -352,26 +352,6 @@ describe('compileModule()', function () {
       expect(m['es-ES']['key']()).toEqual('en-ES');
     });
 
-    it('valid variants', async () => {
-      const mf = new MessageFormat('*');
-      const mp = {
-        'en-US': {
-          key: 'en-US'
-        },
-        'es-MX': {
-          key: 'en-MX'
-        },
-        'es-ES': {
-          key: 'en-ES'
-        }
-      };
-
-      const m = await getModule(mf, mp);
-      expect(m['en-US']['key']()).toEqual('en-US');
-      expect(m['es-MX']['key']()).toEqual('en-MX');
-      expect(m['es-ES']['key']()).toEqual('en-ES');
-    });
-
     it('mixed variants', async () => {
       const mf = new MessageFormat('*');
       const mp = {
--- a/packages/core/src/compile-module.test.ts
+++ b/packages/core/src/compile-module.test.ts
@@ -332,77 +332,77 @@ describe('compileModule()', function () {
 
   describe('locale variants', () => {
     it('valid variants', async () => {
-      const mf = new MessageFormat("*");
+      const mf = new MessageFormat('*');
       const mp = {
-        "en-US": {
+        'en-US': {
           key: 'en-US'
         },
-        "es-MX": {
-          key: "en-MX"
+        'es-MX': {
+          key: 'en-MX'
         },
-        "es-ES": {
-          key: "en-ES"
+        'es-ES': {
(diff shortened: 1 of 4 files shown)


packages/core/src/compiler.ts
```diff
@@ -74,7 +74,8 @@ export default class Compiler {
     if (typeof src === 'object') {
       const result: StringStructure = {};
       for (const key of Object.keys(src)) {
-        const pl = (plurals && plurals[key]) || plural;
+        const normalizedPluralsKey = key.length > 2 ? normalize(key) : key;
+        const pl = (plurals && plurals[normalizedPluralsKey])  || plural;
```
dpchamps, line 78: I considered initially making this a double fallback:

```
const pl = (plurals && plurals[normalizedPluralsKey]) || (plurals && plurals[key]) || plural;
```

But could not determine whether this covers a desirable fallback path. Could use guidance here.

dpchamps:
I'm not quite sure how to address the test failures... They seem to be unrelated to my changeset and passing locally under the same node versions, is there known flakyness here?

packages/core/src/compile-module.test.ts
```diff
@@ -329,4 +329,80 @@ describe('compileModule()', function () {
...
+      };
+
+      const m = await getModule(mf, mp);
+
+      expect(m['en-US']['key']()).toEqual('en-US');
+      expect(m['es-MX']['key']()).toEqual('en-MX');
+      expect(m['es-ES']['key']()).toEqual('en-ES');
+    });
+
+    it('valid variants', async () => {
```
cdaringe, line 355: dupe `it` spec names. may consider making them unique

cdaringe: ah, the whole test case is duped :)

dpchamps: was just a duplicate spec, fixed now

packages/core/src/compiler.ts
```diff
@@ -74,7 +74,8 @@ export default class Compiler {
     if (typeof src === 'object') {
       const result: StringStructure = {};
       for (const key of Object.keys(src)) {
-        const pl = (plurals && plurals[key]) || plural;
+        const normalizedPluralsKey = key.length > 2 ? normalize(key) : key;
```
cdaringe, line 77: could we just `normalize` 💯 % of the time? seems like this length check is already implemented there. may be nice to dedupe it

dpchamps: `normalize` will throw if given unexpected input, so no.

cdaringe (approved):

packages/core/src/compiler.ts
cdaringe: if 1) normalization is to be trusted, and 2) only normalized values are to be supported, then this appears to be a safe refactor

packages/date-skeleton/src/get-date-formatter.test.ts
```diff
@@ -120,7 +120,8 @@ describe('Options', () => {
     expect([
       '2 Dhuʻl-Hijjah 1426',
       'Dhuʻl-Hijjah 2, 1426',
-      'Dhuʻl-Hijjah 2, 1426 AH'
+      'Dhuʻl-Hijjah 2, 1426 AH',
+      '2 Dhuʻl-Hijjah 1426 AH'
```
eemeli, line 124: Thank you, this looks like the right fix for this test. Would you be ok if I cherry-picked this commit separately from the rest of the PR?

eemeli (changes requested):
This looks like it may be a breaking change. Consider this case, and how it works at the moment:

```js
// input
const  mf = new MessageFormat(['en', 'no']) // English & Norwegian
compileModule(mf, { 'no-one': '{var, plural, one{one} other{other}}' })
```

```js
// output
import { en } from "@messageformat/runtime/lib/cardinals";
import { plural } from "@messageformat/runtime";
export default {
  "no-one": (d) => plural(d["var"], 0, en, { one: "one", other: "other" })
}
```

With normalized as well as exact keys getting tested, that message would end up using Norwegian rather than English pluralisation rules. That's a breaking change.

The current documented behaviour is that:

> If the `messageformat` instance has been initialized with support for more than one locale, using a key that matches the locale's identifier at any depth of a `messages` object will set its child elements to use that locale.

which in your case ends up interacting with the `'*'` locale identifier:
> If `locale` has the special value `'*'`, it will match **all** available locales. This may be useful if you want your messages to be completely determined by your data, but may provide surprising results if your input message object includes any 2-3 character keys that are not locale identifiers.

The documentation bug here is that the latter part refers to "locale identifiers" and only implicitly specifies that to really mean "BCP-47 primary language subtag".

While I agree that we ought to find a solution for your use case, I think this change shouldn't be it.

packages/date-skeleton/src/get-date-formatter.test.ts
dpchamps: please do :)

dpchamps:
@eemeli Yes, I was concerned that there may be some edge cases like this. Makes total sense.

Ideas for how we might be able to find a solution that works without a breaking change:

1. implement a [canonicalization](https://unicode-org.github.io/icu/userguide/locale/#canonicalization) procedure (level 1 should be sufficient)
    a. Alternatively, just extract the key without the canonicalization procedure via `normalize`
3. determine if country code / variant is valid 
4. If so, select the locale identifier from the parsed input
5. (optionally) put this behind a configuration flag, something like `supportLocaleKeyVariants`


That way, by ensuring that the country code / variant is in the set of valid codes if would be very difficult to accidentally specify a key that would be misinterpreted by a variant. 

Supporting a configuration flag would prevent any accidents at all from happening.

dpchamps:
Hey @eemeli was wondering if you had any thoughts wrt https://github.com/messageformat/messageformat/pull/360#issuecomment-1053639087

Apologies for the bump. We have some features that are currently blocked by this and I'm more than happy to contribute to the codebase if it makes sense. If not, totally understandable-- I can find a workaround in the interim. Thanks so much!

cdaringe:
hey @eemeli, hope all is well. re-reading thru this issue i'm still trying to degrok things, but curious if you'd be willing to give @dpchamps or myself any concrete guidance here for a solution you'd be interested in supporting. Thanks for your time and consideration

eemeli:
Closing this in favour of #386.

Issue #361, opened 2022-02-25
bug: Potentially unexpected output given the "*" operator


dpchamps:
While authoring https://github.com/messageformat/messageformat/pull/360, I noticed a potentially unexpected edge case while using the `*` operator:

Let's say I'm authoring some messages, and I want to be overly terse:

```
const messagePacks = {
    "en": {
      es: "Estimation Date: {date, date, ::EEEMMMd}",
      fi: "Financial Incentive Date: {date, date, ::EEEMMMd}",
    },
  };
```

If I were to compile this against the `new MessageFormat("*")` invocation, I'm going to get _very_ unexpected output. Namely, spanish and finnish localized dates.

See the repro here: https://github.com/dpchamps/mf-locale-collision-repro

Idea for a fix: we could add an optional arg to `compile` to indicate `isRecursing`, and then not pull from the plurals at all.

eemeli:
Yeah, this feels like a bug, and rather unlikely to affect real-world users in a way that anyone would (should?) depend on.

PR #362, opened 2022-02-25
fix(core): Do not select plurals for non-top level compilations

dpchamps:
Fixes: https://github.com/messageformat/messageformat/issues/361

This is a sketch to fix 361. I can see it being desirable to hide the inner recursive parameter by creating an "outer" and "inner" function, but I opted for the most simple approach.

### DIFF
--- a/packages/core/src/compiler.ts
+++ b/packages/core/src/compiler.ts
@@ -65,17 +65,21 @@ export default class Compiler {
    * @param src - The source for which the JS code should be generated
    * @param plural - The default locale
    * @param plurals - A map of pluralization keys for all available locales
+   * @param isDescendentCall - Indicates whether or not this is the top level of the recursive call
    */
   compile(
     src: string | StringStructure,
     plural: PluralObject,
-    plurals?: { [key: string]: PluralObject }
+    plurals?: { [key: string]: PluralObject },
+    isDescendentCall?: boolean
   ) {
     if (typeof src === 'object') {
       const result: StringStructure = {};
       for (const key of Object.keys(src)) {
-        const pl = (plurals && plurals[key]) || plural;
-        result[key] = this.compile(src[key], pl, plurals);
+        const pl = isDescendentCall
+          ? plural
+          : (plurals && plurals[key]) || plural;
+        result[key] = this.compile(src[key], pl, plurals, true);
       }
       return result;
     }
--- a/packages/core/src/compile-module.test.ts
+++ b/packages/core/src/compile-module.test.ts
@@ -329,4 +329,23 @@ describe('compileModule()', function () {
       );
     });
   });
+
+  describe('issue 361 regression', () => {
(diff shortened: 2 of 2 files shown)


packages/core/src/compiler.ts
```diff
@@ -65,17 +65,21 @@ export default class Compiler {
...
   ) {
     if (typeof src === 'object') {
       const result: StringStructure = {};
       for (const key of Object.keys(src)) {
-        const pl = (plurals && plurals[key]) || plural;
-        result[key] = this.compile(src[key], pl, plurals);
+        const pl = isDescendentCall
+          ? plural
+          : (plurals && plurals[key]) || plural;
+        result[key] = this.compile(src[key], pl, plurals, true);
```
eemeli, line 82:
```suggestion
        const pk = plurals && plurals[key];
        result[key] = pk
          ? this.compile(src[key], pk)
          : this.compile(src[key], pl, plurals);
```

Wouldn't that have the same effect, with no need for an additional argument?

eemeli (changes requested):
Looks promising! Recommended a slightly alternative approach below. The docs may need a small update as well?

packages/core/src/compiler.ts
dpchamps: Hi, sorry for the delayed response! 

Yes, I believe this would work. It's slightly different in that it supports locale keys at any depth... I didn't realize at first, but your feedback on the other PR leads me to believe that this is a documented feature?

In other words, should the following specification pass?

```
    it('should produce the correct output given nested sub keys that look like plurals', () => {
      const mf = new MessageFormat('*');

      const messagePacks = {
        someGrouping: {
          es: {
            en: 'Estimation Date: {date, date, ::EEEMMMd}',
            fi: 'Financial Incentive Date: {date, date, ::EEEMMMd}'
          }
        },
      };

      const result = compileModule(mf, messagePacks);

      expect(result).toMatch(`new Intl.DateTimeFormat("es", opt);`);

      expect(result).not.toMatch(`new Intl.DateTimeFormat("fi", opt);`);
      expect(result).not.toMatch(`new Intl.DateTimeFormat("en", opt);`);
    });
```

If yes then we should take your suggestions.

eemeli: @dpchamps Sorry, completely dropped the ball on this. Yes, locale keys may be set at any depth. For instance, you could have a structure like this, and it should work:
```js
{ foo: { bar: { en: 'Bar in English', fi: 'Baari suomeksi' } } }
```

dpchamps: @eemeli Now I've gotta apologize! To be quite honest, I don't remember the context for this fix, so I'll have to set aside some time for focus. 

Are you still interested in landing this patch? If so, I can take a look sometime this week.

eemeli: I'd still be interested in landing something like this, and would be happy to review an update. No hurry on my part though.

PR #364, opened 2022-03-17
docs(guide): Update link for DateFormat skeleton string

floratmin:
Updated link to new ICU website


floratmin:
Can not fix EasyCLA will recommit

PR #366, opened 2022-04-03
Rounding modes

floratmin:
…de-half-floor

Added the rounding modes `Half Ceiling` and `Half Floor`. See [ICU Rounding Modes](https://unicode-org.github.io/icu/userguide/format_parse/numbers/rounding-modes.html)

Did not add rounding mode `Half Odd` because there is no equivalence in `Intl.NumberFormat` and according to their discussion this mode is very unusual and was not included in their specification.


floratmin:
Can not fix EasyCLA will recommit

PR #368, opened 2022-04-03
feat(number-skeleton): Add rounding-modes Half Ceiling and Half Floor

floratmin:
Added the rounding modes `Half Ceiling` and `Half Floor`. See [ICU Rounding Modes](https://unicode-org.github.io/icu/userguide/format_parse/numbers/rounding-modes.html)

Did not add rounding mode `Half Odd` because there is no equivalence in `Intl.NumberFormat` and according to their discussion this mode is very unusual and was especially not included in their specification.

### DIFF
--- a/packages/number-skeleton/src/skeleton-parser/skeleton-parser.test.ts
+++ b/packages/number-skeleton/src/skeleton-parser/skeleton-parser.test.ts
@@ -176,7 +176,11 @@ const tests: { [testSet: string]: { [src: string]: Skeleton } } = {
     'rounding-mode-ceiling': { roundingMode: 'rounding-mode-ceiling' },
     'rounding-mode-down': { roundingMode: 'rounding-mode-down' },
     'rounding-mode-floor': { roundingMode: 'rounding-mode-floor' },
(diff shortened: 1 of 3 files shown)


eemeli (approved):

eemeli:
This looks good! Could you rebase the PR (or merge from `master`) so that we can verify that the tests pass?

floratmin:
I am not sure if `rounding-mode-half-odd` should be added. If the main consideration is to be as close to ICU as possible, then it should also be included. Otherwise, I will merge from master.

eemeli:
Ah, good point. I had not remembered that at the time when I originally wrote this (Nov 2019), rounding modes weren't even included in the Intl.NumberFormat v3 proposal, so setting any value for it will fail:

https://github.com/messageformat/messageformat/blob/fd93870feff6475e37c99f7f7d88a39b3eb7495f/packages/number-skeleton/src/numberformat/options.ts#L234

So the right thing to do here would be to include `rounding-mode-half-odd` as a "valid" option value, and then separately from this PR fix it so that the now-supported rounding modes are actually usable.

floratmin:
Added rounding mode `Half Odd` and reopened pull request from `master` branch.

PR #372, opened 2022-04-12
Dependecy updates

eemeli:
- BREAKING CHANGE for cli, due to updated Node.js requirements (`>= 14`)
- Major version updates:
  - glob 8
  - jest 27
  - make-plural 7
  - react & react-dom 18
  - yaml 2
- Add `test-utils` package, drop `common-tags` dependency (The added package intentionally does not build .js files, as it is meant only for use in tests)
- Refresh root lockfile
- Docs:
  - Update `github-pages` gem
  - Add explicit `webrick` dependency as workaround for ruby v3 support


SlexAxton:
💕💕💕

Issue #328, opened 2021-06-28
Compile only a single translation


jantimon:
I really like the idea of precompiling translation keys during build time 👍

The current API is really cool:

```js
const MessageFormat = require("@messageformat/core");
const compileModule = require("@messageformat/core/compile-module");

const messageFormat = new MessageFormat("en");

console.log(
  compileModule(messageFormat, {
    productCount:
      "{ count, plural,\n   =0 {{formattedCount} Products}\n  one {{formattedCount} Product}\nother {{formattedCount} Products} } ",
  })
);
```

it generates:

```js
import { en } from "@messageformat/runtime/lib/cardinals";
import { plural } from "@messageformat/runtime";
export default {
  productCount: (d) => plural(d.count, 0, en, { "0": d.formattedCount + " Products", one: d.formattedCount + " Product", other: d.formattedCount + " Products" }) + " "
}
```

Unfortunately we would prefer to compile only a single translation so we would like to get the following output:

```js
import { en } from "@messageformat/runtime/lib/cardinals";
import { plural } from "@messageformat/runtime";
export default (d) => plural(d.count, 0, en, { "0": d.formattedCount + " Products", one: d.formattedCount + " Product", other: d.formattedCount + " Products" }) + " "
```

Or even more complicated: 

```js
import { en } from "@messageformat/runtime/lib/cardinals";
import { de } from "@messageformat/runtime/lib/cardinals";
import { plural } from "@messageformat/runtime";
export default ([
  (d) => plural(d.count, 0, en, { "0": d.formattedCount + " Products", one: d.formattedCount + " Product", other: d.formattedCount + " Products" }) + " ",
  (d) => plural(d.count, 0, de, { "0": d.formattedCount + " Produkte", one: d.formattedCount + " Produkt", other: d.formattedCount + " Produkte" }) + " "
])
```

Would it be possible to provide 3 public functions e.g. with the following api?

```js
getCardinalImport('en') -> // import { en } from "@messageformat/runtime/lib/cardinals";
getRuntimeImport() -> // import { plural } from "@messageformat/runtime";
getMessageFormater(messageFormat, "{ count, plural,\n   =0 {{formattedCount} Products}\n  one {{formattedCount} Product}\nother {{formattedCount} Products} } ") // -> (d) => plural(d.count, 0, en, { "0": d.formattedCount + " Products", one: d.formattedCount + " Product", other: d.formattedCount + " Products" }) + " "
```

jantimon:
[...]

PR #386, opened 2023-02-10
Add localeCodeFromKey option

### LINKED ISSUE #359: bug: incorrect locales used in in generated date functions

eemeli:
Fixes #359 
See also #360

Ping @cdaringe, @dpchamps, would this approach work for you?

### DIFF
--- a/packages/core/src/messageformat.ts
+++ b/packages/core/src/messageformat.ts
@@ -72,15 +72,17 @@ export interface MessageFormatOptions<
 
   /**
    * If defined, used by {@link compileModule} to identify and map keys to
-   * the locale identifiers used by plural rules.
+   * the locale identifiers used by formatters and plural rules.
+   * The values returned by the function should match the `locale` argument.
    *
    * Default: `undefined`
    *
    * @example
    * ```js
    * // Support all recognised Unicode locale identifiers
-   * function pluralCodeFromKey(key) {
+   * function localeCodeFromKey(key) {
    *   try {
+   *     // Ignore all language subtags
    *     return new Intl.Locale(key).language
    *   } catch {
    *     return null
@@ -88,7 +90,7 @@ export interface MessageFormatOptions<
    * }
    * ```
    */
-  pluralCodeFromKey?: ((key: string) => string | null | undefined) | null;
+  localeCodeFromKey?: ((key: string) => string | null | undefined) | null;
 
   /**
    * Require all message arguments to be set with a defined value
@@ -220,7 +222,7 @@ export default class MessageFormat<
         biDiSupport: false,
         currency: 'USD',
         customFormatters: {},
-        pluralCodeFromKey: null,
+        localeCodeFromKey: null,
         requireAllArguments: false,
         returnType: 'string',
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/core/src/messageformat.ts
+++ b/packages/core/src/messageformat.ts
@@ -71,6 +71,26 @@ export interface MessageFormatOptions<
   };
 
   /**
+   * If defined, used by {@link compileModule} to identify and map keys to
+   * the locale identifiers used by plural rules.
+   *
+   * Default: `undefined`
+   *
+   * @example
+   * ```js
+   * // Support all recognised Unicode locale identifiers
+   * function pluralCodeFromKey(key) {
+   *   try {
+   *     return new Intl.Locale(key).language
+   *   } catch {
+   *     return null
+   *   }
+   * }
+   * ```
+   */
+  pluralCodeFromKey?: ((key: string) => string | null | undefined) | null;
+
+  /**
    * Require all message arguments to be set with a defined value
    *
    * Default: `false`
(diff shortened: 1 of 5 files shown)


eemeli:
Ping @jantimon, I think this would also fix #328?

```js
// input
const pluralCodeFromKey = key => key === '0' ? 'en' : key === '1' ? 'de' : null
const mf = new MessageFormat(['en', 'de'], { pluralCodeFromKey })
compileModule(mf, ['{count, plural, other{EN}}', '{count, plural, other{DE}}'])
```

```js
// output
import { de, en } from "@messageformat/runtime/lib/cardinals";
import { plural } from "@messageformat/runtime";
export default {
  "0": (d) => plural(d.count, 0, en, { other: "EN" }),
  "1": (d) => plural(d.count, 0, de, { other: "DE" })
}
```

dpchamps:
Changes look good! I like this more explicit arbitrary mapping strategy as opposed to assuming normalization. Thanks!

dpchamps (approved):

packages/core/src/messageformat.ts
```diff
@@ -70,6 +70,26 @@ export interface MessageFormatOptions<
...
+   * function pluralCodeFromKey(key) {
+   *   try {
+   *     return new Intl.Locale(key).language
+   *   } catch {
+   *     return null
+   *   }
+   * }
+   * ```
+   */
+  pluralCodeFromKey?: ((key: string) => string | null | undefined) | null;
```
cdaringe, line 91: I propose `getLocaleCodeFromKey` as an improved name.

We are indexing into a `plurals[localeCode]` internally--indeed! However my original ticket/failure had nothing to do with plurals, thus I was confused why I had to provide a `pluralCodeFromKey`. "To fix my ICU date formatting issue, why would I need to meddle with plurals?"

I studied the source, and now understand. `plurals` is _indexed_ by locale, and ...locale is kind of a superset of pluralCode. To hinge all locale formatting off of the plural code is a fine choice, but users of the MessageFormat API think in terms of locales, not pluralCodes (an implementation detail), so it would be great to keep that tucked inside.

`getLocaleCodeFromKey` as the interface communicates to me: "Hey you, function author, map over each of your object keys and derive the correct locale for them. By the way, it must be compliant with pluralCodes, even if your ICU message has no plurals in it".

Are you open to this suggestion?

eemeli: The challenge with calling the function a "locale code getter" is that at least with the default plurals, you may want the _input_ to be a locale code like `es-MX` while the _output_ might be `es`.

cdaringe: can you help clarify why that is problematic?

`getLocaleCodeFromKey`

input: `Key`
output: `LocaleCode`

key is an arbitrary description. Per your own design, the user controls what type of content the key is. `es` & `es-MX`--both valid locale codes, contingent on what a "code" is defined as. So long as the return goes through `plurals::normalize(...)`, shouldn't it be safe? Perhaps it's not being normalized. Perhaps we should always normalize the return value from this function

cdaringe: hey @eemeli, did you see my feedback?

let me know if you want me to send a patch. otherwise, we can just take the patches as you have authored them. i'm eager to get the next release :)

thanks for your time & energy!

cdaringe: hey @eemeli, sorry to pester! i know you're busy. i'd rather cut a release with what you have than drag out anything further. please kindly let me know if there's anything i can do to help the solution proceed :)

eemeli:
Been punting this until "later" for a while, but ultimately I agree with the option rename suggested by @cdaringe. It wasn't so much the shape of the change that was tricky here, but the scope of what it _doesn't_ do.

Specifically, this PR doesn't actually fix the exact issue that was originally reported in #359, which used the `'*'` locale and locale codes that include a region specifier. For usage like that, the best that's available is to set something like
```js
const localeCodeFromKey = (key) => key === 'es-MX' ? 'es' : null
```
and then you'd get `es` rather than `es-MX` formatting.

 However, the issue _is_ fixed for language+region locale codes if the input list of locales is explicit, i.e. something like `['en-us','es-MX']`.

If someone's interested in improving the former without breaking the API, then I'd be willing to consider a PR on it. But for now, I'm going to add the option as `localeCodeFromKey`, and package a new release with it.

PR #388, opened 2023-04-06
fix(cli): Support Windows paths as directory path inputs

### LINKED ISSUE #387: @messageformat/cli: Using directory path as input does not work on Windows
On Windows environment, you cannot use directories as input path as the version of Glob dependency (v8+) used by CLI supports only forward slashes in paths and `path.join` converts given input to use backward slash that results `glob.sync` returning empty array.

Karolain:
Fixes #387

Backward-slashes are replaced with forward-slashes in glob pattern when using `windowsPathsNoEscape: true` option in `glob.sync`. This does mean that symbols cant be escaped in glob patterns but I think it isn't a problem in this use-case.

Did manual testing with windows and mac and both seemed to still work.

### DIFF
--- a/packages/cli/lib/read-input.js
+++ b/packages/cli/lib/read-input.js
@@ -12,7 +12,10 @@ function listFiles(include, extensions) {
     if (!fs.existsSync(fn)) throw new Error(`Input file not found: ${fn}`);
     if (fs.statSync(fn).isDirectory()) {
       extensions.forEach(ext => {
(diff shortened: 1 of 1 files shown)


eemeli (approved):
Thank you! This looks like it should fix the issue, and I agree that it's really rather unlikely that `*` or `?` are used by anyone as actual path characters with this CLI.

eemeli:
Released as 4.0.1.

PR #396, opened 2023-06-05
Add MF2 resource parser

eemeli:
Add a CST parser for MF2 resources, as defined [here](https://github.com/eemeli/message-resource-wg/blob/main/resource.abnf).

Defines a function with this signature:

```ts
import { parseCST } from '@messageformat/resource';

/**
 * Parse input into a MessageResource CST.
 * Should never throw an Error.
 *
 * @param source - The full source being parsed
 * @param onError - Error handler, may be called multiple times for bad input.
 */
function parseCST(
  source: string,
  onError: (range: CST.Range, msg: string) => void
): CST.Resource
```

The resulting `CST.Resource` corresponds closely to the ABNF, and represents all of its given input. Errors are emitted via a side channel, without stopping the parse.

For entry and section identifiers, their full `string[]` representation is parsed, and errors emitted for duplicates. Declaring a shorter id after having already done so for a matching longer one is considered an error, i.e. a message `a.b` always needs to be defined before `a.b.c`.

Entry values include a `string` value that does not trim their leading whitespace or parse the character escapes, as handling those in the  message parser allows for source positions to be accurately tracked. To wit, a resource mode is added to `parseMessage(source, options?)`, which changes leading-whitespace and escape rules to match messages in resources.


eemeli:
Rebased to leave out the `mf2-messageformat` refactor, that's now in c0339be.

Issue #393, opened 2023-05-19
If plurals has incompatible with a locale keys, all translations in the app break


bigcreate:
we are updating the `messageformat` to v3+ in our applications, but faced with an issue that the whole translation in the app is failed because some plural keys are necessary, and some are redundant in different languages. It's hard to check and fix all locales and all translation files for each locale, besides, it's hard to teach translators to write translations in certain formats and with certain plural keys.
Could you add any backdoor to make the compiler/parser not so strict?
Or could you suggest anything about how it might be solved on our side?

eemeli:
Could you give an example of what you mean? Only the `other` key is required for plurals.

bigcreate:
@eemeli , for example, if for German we have `many` key, it fails
and yes, I was wrong about the required keys, I've thought that, for example, for Polish `one`, `few`, `many`, `other` are required, but it works without `many`, for example
So only redundant keys  like `many` for German lead to the error

eemeli:
Ok, that makes sense. Really you ought to be able to warn and filter out those before formatting, but adding a default-`true` option like `strictPluralKeys` could be done. A PR on this would be welcome.

bigcreate:
Hi @eemeli, could we expect that the issue might be resolved on your side? Unfortunately, we are fully packed with our own tasks. And to be honest, these breaking changes weren't in our plans, and sooner or later they will break our apps (we are not using yet v3 in prod)

eemeli:
To be honest as well, this has rather low priority for me atm, and I might or might not get to fixing it myself sometime in the next few months. If you'd like to discuss a support contract, please reach out to me directly at eemeli@gmail.com.

eemeli:
Fixed in #397 and #400.

PR #397, opened 2023-06-06
feat(parser): Introduce new ParseOption to disable strict plural key check

jaideepsingh:
Issue #393 

As recommended in the issue referenced above, this change introduces a new option to the parser: `strictPluralKeys`, which is set to `true` by default. Relevant unit tests have also been added.

Once the PR is merged and a new version of `@messageformat/parser` released, I'll create a subsequent PR to add support for this option in `@messageformat/core`.

### DIFF
--- a/packages/parser/src/parser.test.ts
+++ b/packages/parser/src/parser.test.ts
@@ -593,15 +593,29 @@ describe('Errors', () => {
     }).toThrow();
   });
 
-  it('should not allow invalid keys for plurals', function () {
-    expect(function () {
-      parse('{NUM, plural, one { 1 } invalid { error } other { 2 }}');
-    }).toThrow();
-    expect(function () {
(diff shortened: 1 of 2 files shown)


jaideepsingh:
> This looks good! A couple of nitpicks inline.

@eemeli Thank you for checking it out. The suggestions look perfect; I've added them.

jaideepsingh:
@eemeli Thanks for the merge. Would it be possible to release this change as a new minor version for `@messageformat/parser`?

eemeli:
I'd prefer releasing both this and your follow-up update to `@messageformat/core` at the same time. Having also the `core` change merged would help validate that this `parser` API is doing what it's meant to do. It's a monorepo, so working on `core` will depend on the in-tree `parser`.

PR #400, opened 2023-06-07
feat(core): Add `strictPluralKeys` option for supporting the corresponding option in parser

jaideepsingh:
Issue #393 

As a part of the above mentioned issue, this change adds the `strictPluralKeys` option to the `@messageformat/core` package.

The corresponding option has already been added to `@messageformat/parser` in #397.

### DIFF
--- a/packages/core/src/messageformat.ts
+++ b/packages/core/src/messageformat.ts
@@ -121,8 +121,8 @@ export interface MessageFormatOptions<
   /**
    * Enable strict checks for plural keys according to
    * {@link http://cldr.unicode.org/index/cldr-spec/plural-rules | Unicode CLDR}.
-   * When set to `false`, the compiler will also accept any invalid plural keys. Also see the
-   * {@link http://messageformat.github.io/messageformat/api/parser.parseoptions.strictPluralKeys/ | parser option}.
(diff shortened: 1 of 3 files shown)


packages/core/src/messageformat.ts
```diff
@@ -117,6 +117,16 @@ export interface MessageFormatOptions<
    * Default: `false`
    */
   strict?: boolean;
+
+  /**
+   * Enable strict checks for plural keys according to
+   * {@link http://cldr.unicode.org/index/cldr-spec/plural-rules | Unicode CLDR}.
+   * When set to `false`, the compiler will also accept any invalid plural keys. Also see the
+   * {@link http://messageformat.github.io/messageformat/api/parser.parseoptions.strictPluralKeys/ | parser option}.
```
eemeli, line 125:
```suggestion
   * When set to `false`, the compiler will also accept any invalid plural keys.
   * Also see the corresponding {@link @messageformat/parser#ParseOptions | parser option}.
```

eemeli (approved):
A couple of small fixes, but this too looks good!

jaideepsingh:
@eemeli Thanks for your inputs. I've included them.

eemeli:
New versions released with this option: https://github.com/messageformat/messageformat/releases

PR #403, opened 2023-06-13
feat: support locale-specific customFormatters

### LINKED ISSUE #402: RFC: Support customFormatters locale specific modules
# Problem

The current custom formatter API does not allow me to import my runtime module with locale awareness.

Current: 
```js
export const phone: CustomFormatter = {
  // ...snip
  module: '@messageformat/phone'
};
```

Desired:
```
export const phone: CustomFormatter = {
  // ...snip
  module: (localeCode) => `@messageformat/upcase/${localeCode}` // e.g. "@messageformat/upcase/mx"
};
```

# Discussion

In this respect, the generated ESM javascript can use much smaller implementations for the compilation of interest!

cdaringe:
# Problem Statement

Given a locale & a message with a custom formatter requirement, I want my compiled output to use a formatter tailored to that locale, but cannot due to limitations in the current API.

Closes #402

## Motivating Context

Presume that some custom formatter is expensive to load for _all_ supported locales.

In the current API, my formatter must do something like this:

```js
formatter: (value, locale) => 
  isUs(locale)
  ? formatUs(value)
  : isMx(locale)
  ? formatMx(value)
//  ...SNIP N locales ...
  : isFinland(value)
  ? formatFn(value)
  : value
```

```js
// compiled-output.js
import { fmt } from "formatters/formatter-that-knows-about-all-locales-under-the-sun"
export const msg = d => fmt(d.foo)
```

 Presume that each formatting function is **expensive, say 1kb**. Now my app must install (N-1) kb of wasted formatters on each application load.

# Solution

Support locale-aware custom formatters.

```js
// compiled-output.js
import { fmt as fmt_enUS } from "formatters/en-US"
export const msg = d => fmt_enUS(d.foo)
```

```js
// compiled-output.js
import { fmt as fmt_esMX } from "formatters/en-MX"
export const msg = d => fmt_enMX(d.foo)
```

In the new state, customFormatters do not need to bundle _all_ locale concerns into a single module.


packages/core/package.json
```diff
@@ -13,7 +13,8 @@
   ],
   "contributors": [
     "Alex Sexton <alexsexton@gmail.com>",
-    "Eemeli Aro <eemeli@gmail.com>"
+    "Eemeli Aro <eemeli@gmail.com>",
+    "cdaringe<cdaringe@gmail.com>"
```
cdaringe, line 17: is that bad form 😄 ?

packages/core/src/compiler.ts
```diff
@@ -182,16 +195,20 @@ export default class Compiler {
...
+        }
         if (token.param) {
           if (pluralToken && this.options.strict) pluralToken = null;
           const arg = this.getFormatterArg(token, pluralToken);
           if (arg) args.push(arg);
         }
-        fn = token.key;
-        this.setFormatter(fn);
+        fn = isLocaleSpecificFormattingModule
+          ? `${token.key}__${this.currentLocaleCode?.replace(/[^a-zA-Z]/g, '')}`
```
cdaringe, line 207: e.g. `es-US` => `esUS`

packages/core/src/compiler.ts
```diff
@@ -303,26 +320,50 @@ export default class Compiler {
     }
   }
 
-  setFormatter(key: string) {
-    if (this.runtimeIncludes(key, 'formatter')) return;
-    let cf = this.options.customFormatters[key];
+  setFormatter(runtimeFormatterKey: string, originalFormatterKey?: string) {
```
cdaringe, line 323: e.g. `setFormatter("phone__esUS", "phone")`

packages/core/src/messageformat.ts
```diff
@@ -61,12 +70,18 @@ export interface MessageFormatOptions<
...
+          formatter: DefaultCustomFormatter;
           arg?: 'string' | 'raw' | 'options';
           id?: string;
           module?: string;
+        }
+      | {
+          formatter: LocaleModuleCustomFormatter;
+          arg?: 'string' | 'raw' | 'options';
+          id?: string;
+          module?: (locale: string) => string;
```
cdaringe, line 84: some of these keys could be deduped, but they were small enough i just dupe'd em (e.g. `arg`, `id`)

packages/core/src/compiler.ts
eemeli: Rather than putting together a new sanitization solution, it'd be better to use [`identifier()`](https://www.npmjs.com/package/safe-identifier#identifierkey-string-unique-boolean-string).

eemeli (commented):
Took a quick glance, but I'll need to find time for a proper review. I'm a bit concerned about the expansion of state during the compilation; the `plural` arg is a bit badly named, but I would've thought that it was already tracking the current locale?

cdaringe:
>  I'm a bit concerned about the expansion of state during the compilation

justifiably so! dig in at your convenience, let me know your thoughts

> the plural arg is a bit badly named, but I would've thought that it was already tracking the current locale?

the `plural/plurals` doesn't track locale. it has a `locale` field, but it's not honest. it's really just the lang, disguised as the locale.

for recollection:

```js
plurals = {
  // Record<LANG_STRING, PluralData>
en: { cardinals: ..., lc: 'en', locale: 'en' }
}
```

Problem is... i _need_ the actual/full locale code, e.g. `en-US`, not just the lang

cdaringe:
Hey @eemeli, i know i'm blowing up your mentions this week 😄 . I'll go silent in all other forums if it helps! Would you be willing to give me some 👀 on this anytime soon?

eemeli:
Hi, sorry for the delay here -- been on vacation (and still am...) & trying to stay afk. Back for a couple of days next week.

Reviewing this PR and how it fits in with the rest, I'm starting to get the sense that there's some premature caching happening with the `'*'` locale. This ends up being related to #361 as well. The current locale _should be_ available within the plural object. See for example the usage here: https://github.com/messageformat/messageformat/blob/04b395220c6f1327ede71fa5c888934ab8648e23/packages/core/src/compiler.ts#L185

Not sure if/when I'll have time for this myself, but for `'*'` the `plurals` array should probably be initialised not with all locales, but with the default locale and some marker indicating the wildcard. Then around here https://github.com/messageformat/messageformat/blob/04b395220c6f1327ede71fa5c888934ab8648e23/packages/core/src/compiler.ts#L81
the lookup should set the locale in the plural object according to the `key`.

cdaringe:
No worries. Please, ignore this, and enjoy your vacation :)

> The current locale should be available within the plural object

Indeed! However, locale is not in actuality available. As I [mention in my comment above](https://github.com/messageformat/messageformat/pull/403#issuecomment-1591605774), `locale` is not actually `locale` at all. It's pretty much just `lang`, which is too lossy.

eemeli:
`plural.locale` is only the primary subtag if you use `*` as the constructor locale; if you provide an array of locale identifiers there, they are used exactly.

This should probably be documented better.

cdaringe:
Hm. Interesting. Shucks, I still feel like that my be a modeling issue in the existing plurals interface? I would ask "is a subtag a locale?". I'd suggest it is _not_. Either way, it's probably worth me learning OR us collaborating on what the correct nomenclature is for `lang + region` unambiguously. Perhaps `PluralObject` warrants some refinement?

> array should probably be initialised not with all locales, but with the default locale ... the lookup should set the locale in the plural object according to the key

i'm ...kinda following, but not 100%. i'm not confident enough to send a patch on this yet. i think you're talking about in the `Compiler::compile` method, maybe creating a new `PluralObject` on the fly with narrowed _stuff_.

If something were to happen with the above, would it be a mandatory precursor to this PR?

eemeli:
> Hm. Interesting. Shucks, I feel like that there may be a modeling issue in the existing plurals interface? I would ask "is a subtag a locale?". I'd suggest it is _not_. Either way, it's probably worth me learning OR us collaborating on what the correct nomenclature is for `lang + region` unambiguously. Perhaps `PluralObject` warrants some refinement?

Not entirely sure what you're asking for here. To clarify, `plural.locale` is a BCP 47 language tag identifying a locale. Such a locale identifier requires at least a language tag, but may also include other subtags, such as a region identifier.

The plural objects are created during the MF constructor call, in two different ways:
1. If the given list of locales is explicit, we get an MF instance supporting exactly those locales.
2. If the user asks for all locales with `'*'`, plural objects are created for all supported _languages_, i.e. locales with only the primary tag (and pt-PT, but that's not relevant here).

The disparity in expectations arises when the second path is taken, _and_ a user uses `localeCodeFromKey` to identify a supported locale code. If that key is something that looks like a BCP 47 locale identifier (e.g. `en-CA` as a key mapping to the `en` locale code), they may expect for that _key_ to have some relevance for custom formatters, but it doesn't. Only the resolved locale code matters.

I don't think there's really a sensible way to get around this, and I'm not sure that it should be fixed. If you're doing complicated things, you should probably not use `'*'` but instead explicitly declare your supported locales.

> > array should probably be initialised not with all locales, but with the default locale ... the lookup should set the locale in the plural object according to the key
> 
> i'm ...kinda following, but not 100%. i'm not confident enough to send a patch on this yet. i think you're talking about in the `Compiler::compile` method, maybe creating a new `PluralObject` on the fly with narrowed _stuff_.
> 
> If something were to happen with the above, would it be a mandatory precursor to this PR?

No, that's optimisation that should be doable entirely separately. The next step with this PR is to get rid of the locale normalisation, additional state around that, and just use `plural.locale`. Also, please drop the docs updates; that gets pushed to the site from the main branch, so needs to wait for a version release.

eemeli (changes requested):
See previous comment.

cdaringe:
😵 Got it! Thanks for setting me straight. I now understand much better and see your perspective.

I've adapted the PR to your request, and it has shrunk up a bit 😄 !

packages/core/src/compiler.ts
```diff
@@ -171,7 +171,12 @@ export default class Compiler {
         this.setRuntimeFn('plural');
         break;
 
-      case 'function':
+      case 'function': {
+        const formatter = this.options.customFormatters?.[token.key];
```
eemeli, line 175: The option is always defined here.
```suggestion
        const formatter = this.options.customFormatters.[token.key];
```

packages/core/src/compiler.ts
```diff
@@ -171,7 +171,12 @@ export default class Compiler {
...
+        const formattingModuleRequest =
+          formatter && 'module' in formatter ? formatter.module : null;
+        const isLocaleSpecificFormattingModule =
+          typeof formattingModuleRequest === 'function';
```
eemeli, line 179: The temporary variable is more confusing than not.
```suggestion
        const isLocaleSpecificFormattingModule = typeof formatter?.module === 'function';
```

Also, "isLocaleSpecificFormattingModule" is an excessively long name for a local variable. With changes below, you can probably drop it rather than renaming it.

packages/core/src/compiler.ts
```diff
@@ -182,16 +187,20 @@ export default class Compiler {
           }
         }
 
-        args.push(JSON.stringify(this.plural.locale));
+        if (!isLocaleSpecificFormattingModule) {
+          args.push(JSON.stringify(this.plural.locale));
+        }
```
eemeli, line 192: This change needs to be reverted. We still want the locale as an argument because we should not assume that each locale will have an independent implementation. It should still be possible to e.g. say that all English (en, en-US, en-CA, en-UK, ...) use the same implementation, but possibly with some region-specific considerations.

packages/core/src/compiler.ts
eemeli: These arguments would be clearer for me with reversed order, i.e. "the `key` of the thing we're importing", and optionally "the `alias` we use for it internally". I'm also not a fan of really long names for local variables, esp. when combined with our 80 char width style.

eemeli (changes requested):
Sorry for the delayed and partial review; too many things on my to-do list atm...

Only got part of the way through the compiler; haven't looked yet at what's happening with the `Object.assign`/`.bind()` stuff or anything further than that.

packages/core/src/compile-module.ts
```diff
@@ -28,6 +28,10 @@ function stringifyRuntime(runtime: RuntimeMap) {
 
   for (const [name, fn] of Object.entries(runtime)) {
     if (fn.module) {
+      /* istanbul ignore next reason: impossible case */
+      if (typeof fn.module !== 'string') {
+        throw new Error(`expected string module name`);
+      }
```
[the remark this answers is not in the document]
cdaringe: type narrowing only

packages/core/src/compiler.ts
cdaringe: 5bf4e84

packages/core/src/compiler.ts
cdaringe: 5bf4e84

cdaringe:
1. locale is now always a param, per discussion
2. long var names compacted, per discussion

> [setFormatter] These arguments would be clearer for me with reversed order

This method is used in many places, and reversing the params would be a breaking change requiring a bulk edit in this module. Give me another review pass now that it's been through a first round of requested edits and assert that you still want this.

packages/core/src/compiler.ts
```diff
@@ -303,21 +309,28 @@ export default class Compiler {
...
     if (this.runtimeIncludes(key, 'formatter')) return;
-    let cf = this.options.customFormatters[key];
+    const cf = this.options.customFormatters[parentKey || key];
     if (cf) {
-      if (typeof cf === 'function') cf = { formatter: cf };
+      const cfo = typeof cf === 'function' ? { formatter: cf } : cf;
+
       this.runtime[key] = Object.assign(
-        cf.formatter,
+        cloneFunction(cfo.formatter),
```
eemeli, line 319: Ok, I see now why mutating the formatter can be a bad idea: When it's not a built-in formatter that we're providing, it means we're modifying a value that a user has provided.

So using `.bind()` here is appropriate. See below for why you ought to inline the code for that, as you had it previously.

packages/core/src/compiler.ts
```diff
@@ -303,21 +309,28 @@ export default class Compiler {
...
+                typeof cf.module === 'function' && this.plural.locale
+                  ? cf.module({ locale: this.plural.locale })
+                  : cf.module
+            }
           : { id: null, module: null }
       );
     } else if (isFormatterKey(key)) {
       this.runtime[key] = Object.assign(
-        Formatters[key],
+        cloneFunction(Formatters[key]),
```
eemeli, line 333: OTOH, here the formatter is an internal one, so we don't need to clone it and can mutate it without ill effect.

packages/core/src/compiler.ts
```diff
@@ -303,21 +309,28 @@ export default class Compiler {
...
       this.runtime[key] = Object.assign(
-        cf.formatter,
+        cloneFunction(cfo.formatter),
         { type: 'formatter' } as const,
         'module' in cf && cf.module && cf.id
-          ? { id: identifier(cf.id), module: cf.module }
+          ? {
+              id: identifier(cf.id),
+              module:
+                typeof cf.module === 'function' && this.plural.locale
```
eemeli, line 325: Checking `this.plural.locale` is incorrect here. If it may be null, then either it needs to be coerced to an empty string, or the function needs to be able to deal with a null input.

Fixing this will let you drop the `compile-module.ts` change.

packages/core/src/messageformat.ts
```diff
@@ -25,12 +25,37 @@ export type MessageFunction<ReturnType extends 'string' | 'values'> = (
...
+/**
+ * A custom formatter function. See
+ * {@link https://messageformat.github.io/messageformat/custom-formatters/ | Custom Formatters}
+ * for more details.
+ *
+ * @public
+ */
+export type CustomFormatter =
+  | DefaultCustomFormatter
+  | LocaleModuleCustomFormatter;
```
eemeli, line 57: As we're now always including the locale code, am I right that the changes to this signature should be rolled back?

packages/core/src/messageformat.ts
```diff
@@ -61,12 +86,18 @@ export interface MessageFormatOptions<
...
+          formatter: DefaultCustomFormatter;
           arg?: 'string' | 'raw' | 'options';
           id?: string;
           module?: string;
+        }
+      | {
+          formatter: LocaleModuleCustomFormatter;
+          arg?: 'string' | 'raw' | 'options';
+          id?: string;
+          module?: (_: { locale: string }) => string;
```
eemeli, line 100: Why does this expect the `locale` to be wrapped within an object? Wouldn't it be simpler to use `(locale: string) => string`?

eemeli (changes requested):
Finally got through to the end. In general I think the approach is good, but we should be able to make the changes here more surgical; see inline comments.

eemeli:
> > [setFormatter] These arguments would be clearer for me with reversed order
> 
> This method is used in many places, and reversing the params would be a breaking change requiring a bulk edit in this module. Give me another review pass now that it's been through a first round of requested edits and assert that you still want this.

Let's return to this after the current round of changes. My main concern is that with separate terms for the local variable name and the import name, those terms ought to correspond to the names used for the JS `import { foo as bar } ...` statement, where this is considered as "aliasing".

packages/core/src/messageformat.ts
cdaringe: I had it that way originally, but i changed it to the object form so it could be extended without being a breaking change

packages/core/src/messageformat.ts
cdaringe: No, otherwise I’m in no better position than when I started without the PR. the function form is precisely what lets me generate locale based module imports. The param is runtime, versus compile time

packages/core/src/messageformat.ts
eemeli: Locale is good to keep as a separate argument. For example, that's the way it is in the MessageFormat constructor. Adding more arguments can be done later.

packages/core/src/messageformat.ts
eemeli: The function form of `module` allows that. This is the signature of the function that'll be called during the formatting. After https://github.com/messageformat/messageformat/pull/403#discussion_r1272317743, it'll always include the locale argument.

packages/core/src/compile-module.ts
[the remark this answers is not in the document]
cdaringe: removed! 72396da

packages/core/src/compiler.ts
```diff
@@ -24,7 +24,7 @@ interface RuntimeEntry {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   (...args: any[]): unknown;
   id?: string | null;
-  module?: string | null;
+  module?: string | ((_: { locale: string }) => string) | null;
```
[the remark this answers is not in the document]
cdaringe: ack. fixed, 72396da

packages/core/src/compiler.ts
cdaringe: ack, fixed 72396da

packages/core/src/messageformat.ts
cdaringe: sure, i can get on board with that. refactored in ee31644

packages/core/src/messageformat.ts
cdaringe: got it. i misinterpreted. my bad. addressed in a681a58

packages/core/src/compiler.ts
```diff
@@ -171,7 +171,11 @@ export default class Compiler {
         this.setRuntimeFn('plural');
         break;
 
-      case 'function':
+      case 'function': {
+        const formatter = this.options.customFormatters[token.key];
+        const formattingModuleRequest =
+          formatter && 'module' in formatter ? formatter.module : null;
+        const isModuleFn = typeof formattingModuleRequest === 'function';
         if (!this.options.customFormatters[token.key]) {
```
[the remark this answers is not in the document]
cdaringe: i wish we could do that, but TS doesn't allow you access an indexed field unless it's narrowed the type to an indexable type. `string` isn't indexable by string, so TS barfs. TS demands that you formatter to a form where `'module'` is guaranteed on an object form of `formatter`, so... no dice. nonetheless, in the spirit of your comment, i condensed it down to a single expression. a681a58

packages/core/src/compiler.ts
cdaringe: it's your call. i am biased to _not mutating the underlying fn_ ever, out of what I perceive to be general purpose best practices, but that is entirely subjective :)

cdaringe:
Alrighty, pass #3, complete! Back to you!

packages/core/src/compiler.ts
```diff
@@ -171,7 +171,12 @@ export default class Compiler {
...
         break;
 
-      case 'function':
+      case 'function': {
+        const formatter = this.options.customFormatters[token.key];
+        const isModuleFn =
+          formatter &&
+          'module' in formatter &&
+          typeof formatter.module === 'function';
         if (!this.options.customFormatters[token.key]) {
```
eemeli, line 180:
```suggestion
        if (!formatter) {
```

packages/core/package.json
eemeli: I don't really mind this, but please do at least match the format of the other entries.

packages/core/src/compiler.ts
eemeli: Let's just mutate it, and inline the `cloneFunction()` above.

eemeli (changes requested):
Nearly there, I think. See inline replies.

packages/core/src/compiler.ts
cdaringe: b542f00

packages/core/src/compiler.ts
cdaringe: done, 9ee2bf1

cdaringe:
pass 4 complete. back to you! thanks for the feedback. i think the PR is in much better shape that it was on first pass

eemeli (approved):
There are still a few code style issues in `setFormatter()`, but we can take care of them separately later.

The docs at `docs/custom-formatters.md` will need to get updated together with the next release. @cdaringe, would you be able/interested to submit a separate PR for that?

cdaringe:
sound good. please see #408

PR #404, opened 2023-06-15
Support negative numbers as unquoted literal values

eemeli:
This is a sketch of an idea for how negative numbers could be supported as unquoted literal values.


eemeli:
Closing; support for unquoted negative numbers is included in #414.

PR #408, opened 2023-08-02
docs: add locale specific formatting modules

cdaringe:
# Problem

#403 requires documentation

# Solution

Add documentation

# Discussion

I will be out-of-office for ~1-2 weeks. Please make edits _directly_ to the PR vs requesting changes, if feasible :)!


eemeli:
I've cleaned up the docs changes a bit and left out the example, as this really ought to finally get published.

PR #414, opened 2024-01-03
Update MF2 to match upstream changes

eemeli:
This should bring the implementation to parity with the [current spec](https://github.com/unicode-org/message-format-wg/tree/2a69d2a17ee7476282bfa824fd4ca62e0a03f6e1/spec), pre-emptively also including changes from:
- https://github.com/unicode-org/message-format-wg/pull/560
- https://github.com/unicode-org/message-format-wg/pull/574
- https://github.com/unicode-org/message-format-wg/pull/577

Some ESLint changes are included, which is why this touches some files outside the `packages/mf2-*` directories. I've also dropped commitlint, as it was getting too annoying during development.

The `@messageformat/fluent` package is here made ES-only, as that was the easiest way to satisfy TS regarding its `@fluent/syntax` dependency. This may be a temporary thing; we'll see. The Fluent -> MF2 converters now use duck-typing to detect plural selectors; this can be disabled with an option.

The XLIFF representation is streamlined a bit. Am considering using `<originalData>` and `<res:resourceData>` for some of what's currently in a wholly custom `<mf:messageformat>` container.

The `messageformat` exports are refactored quite a bit, adding new endpoints `messageformat/functions` and `messageformat/functions/utils` with relevant parts of the API.

Two new data model tools are added:
```js
import { validate, visit } from 'messageformat'
```
`visit` is a specialized message data visitor, and `validate` checks that the model does not have any data model errors, and reports the function & variable dependencies of a message. Currently it effectively duplicates functionality from the CST parser; that should get resolved at some point, but that'll require a datamodel -> CST mapper so that the errors can still be reported by character position.

The JSON test fixtures are refactored a bit, separating out:
- syntax errors
- tests that are independent of all function calls
- tests for each function

The internal file structures of the main package have evolved quite a bit, and feel like they make some sense again. I'm not really expecting anyone to review all the changes here, as that would likely be quite a slog.

CC @aphillips, @mradbourne


aphillips:
**_Very_** happy to see this working going on. I didn't read to the bottom of the code yet. I do notice lots of good general code hygiene in here as well.

eemeli:
Added CST backreferences to the parsed data model, available as `[Symbol.for('CST')]` on most nodes. This let me dedupe the data model error validation, which sparked another refactor of the visitor. It now lets you run code on both node entry & exit, provides more context, has more nodes it covers, and a catch-all `node()` that can be useful for stuff like: https://github.com/messageformat/messageformat/blob/ab0f80715fc592d1b021c61b2520342dbb716b7d/packages/mf2-messageformat/src/stringifier/message.test.ts#L11-L18

eemeli:
Added `stringifyCST()`, as that was clearly missing. This update should now be pretty much complete; next steps can be taken separately.

PR #425, opened 2024-07-25
fix: align mf2 number function to spec

lucacasonato:
https://github.com/tc39/proposal-intl-messageformat/pull/66

### DIFF
--- a/packages/mf2-fluent/src/fluent.test.ts
+++ b/packages/mf2-fluent/src/fluent.test.ts
@@ -437,7 +437,7 @@ describe('formatToParts', () => {
       }
     `;
 
-    const res = fluentToResource(src);
+    const res = fluentToResource(src, "en");
 
     test('defined formatted variable', () => {
       const foo = res.get('foo')?.get('')?.formatToParts({ num: 42 });
@@ -446,7 +446,7 @@ describe('formatToParts', () => {
         {
           type: 'number',
           source: '$num',
-          locale: 'en-US',
+          locale: 'en',
           parts: [{ type: 'integer', value: '42' }]
         }
       ]);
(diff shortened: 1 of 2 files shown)


packages/mf2-messageformat/src/functions/number.ts
```diff
@@ -61,23 +61,18 @@ export function number(
     Intl.PluralRulesOptions & { select?: 'exact' | 'cardinal' | 'ordinal' } = {
     localeMatcher
   };
-  switch (typeof input) {
-    case 'bigint':
-    case 'number':
-      value = input;
```
eemeli, line 67: This is left out in the change, and is necessary for numeric inputs.

eemeli (changes requested):

packages/mf2-messageformat/src/functions/number.ts
lucacasonato: Oh I see - I didn't realize value and input were separate variables. They are not separate in the spec. Maybe I could align them here?

eemeli: Aligning them would be fine, if you can get TS to be satisfied with that...

eemeli: Actually, it might've been not wanting to change an argument value that drove this design, and so changing line 59 to something like
```ts
let value = input;
```
would probably be fine.

lucacasonato: OK will do.

packages/mf2-fluent/src/fluent.test.ts
eemeli, line 1: Why the changes here? Not objecting, they just seem irrelevant?

eemeli (approved):

packages/mf2-fluent/src/fluent.test.ts
lucacasonato: The test did not pass on my `en-DK` system - the test was assuming the system locale is set to `en-US`.

PR #426, opened 2024-07-26
Fix browser tests and express deps correctly

lucacasonato:
Ever since the chai update to 5.x, the browser tests were broken.

Also added some missing top level deps (to allow installing deps with `pnpm i` and `deno install`), and cleaned up the babel config.


babel.config.js
```diff
@@ -1,36 +1,7 @@
-// Because we need `allowDeclareFields: true`, we can't use
-// @babel/preset-typescript as it'd get run _after_ the class-properties
-// plugin, so we mess around like this. Once we can drop Node.js 10 support,
-// we can also drop the class-properties plugin, and that'll allow for
-// simplifying this.
```
lucacasonato, line 5: `class-properties` is in `@babel/preset-env` now

eemeli (approved):
Thank you! I tweaked the test versions a bit to match the browserslist query, and what's available on Browserstack. But with this it looks like the browser tests are actually completing once again.

PR #457, opened 2025-08-31
fix: make the data-model data

fatbrain:
Remove the use of `Map` from data-model, to enable using pre-parsed messages. Removing the need for parsing while running my application.

### DIFF
--- a/mf2/fluent/src/fluent-to-message.ts
+++ b/mf2/fluent/src/fluent-to-message.ts
@@ -129,18 +129,18 @@ function asExpression(exp: Fluent.Expression): MF.Expression {
         throw new Error(`More than one positional argument is not supported.`);
       }
       if (named.length > 0) {
-        annotation.options = new Map();
+        annotation.options = {}
         for (const { name, value } of named) {
           const quoted = value.type !== 'NumberLiteral';
           const litValue = quoted ? value.parse().value : value.value;
-          annotation.options.set(name.name, {
+          annotation.options[name.name] = {
             type: 'literal',
             value: litValue
-          });
+          };
         }
       }
       if (annotation.name === 'number') {
-        const style = annotation.options?.get('style');
+        const style = annotation.options?.style;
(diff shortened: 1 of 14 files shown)


eemeli:
Could you explain a bit more how the current approach is making things difficult for you, preferably with some code examples? It seems like this might be best considered first via an issue, rather than jumping directly into an implementation.

eemeli:
Splitting the parsing like that isn't something that's been considered very deeply here, as the original design of the library is meant to act as a polyfill for a native-code `Intl.MessageFormat` implementation, where we can expect that parsing a syntax string is comparable to parsing JSON -- but your point is valid, esp. as we consider workflows where a non-MF2 syntax is parsed into a JSON structure in a back-end.

Using Objects rather than Maps would also match the spec's [`message.json`](https://github.com/unicode-org/message-format-wg/blame/main/spec/data-model/message.json) definition of the data model, rather than its [TypeScript description](https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/README.md), so it ought to be supported at least for input.

The question then becomes, is there a reason to support both Maps and Objects, or only Objects? At the very least, we should make sure that prototype pollution is not an issue.

fatbrain:
If prototype pollution is handled, I don't see a reason to also support `Map`. But then again I've not considered _mf2_ in my reasoning. Is the `Intl.MessageFormat` suppose to also support `MessageData`, if so, I don't think I've seen `Map` being part  of any of the other `Intl.XXX` interfaces?

If I wanted to add a short helper function in order to safely get a own-property from an object, you have any suggestion where to put it, or you think it's something that can be handled inline for each case?

Something like:

```ts
const getOwn = <T extends object>(object: T | undefined, name: keyof T) =>
  object && Object.hasOwn(object, name) ? object[name] : undefined;
```

Any and all input greatly appreciated, and thanks for such quick responses.

eemeli:
@fatbrain Looks like a Prettier formatter run is needed?

PR #470, opened 2026-04-23
Use `pgs:switch` & `pgs:case` instead of `mf:select` for MF2 Xliff representation

### LINKED ISSUE #456: Xliff: use xliff 2.2 schema for plural and gender
https://docs.oasis-open.org/xliff/xliff-core/v2.2/xliff-extended-v2.2-part2.html#plural_gender_select_module

See 4.9.6 with examples of- this could be good to leverage

eemeli:
Fixes #456

This adopts the attributes introduced in the [Plural, Gender, and Select Module](https://docs.oasis-open.org/xliff/xliff-core/v2.2/xliff-extended-v2.2-part2.html#plural_gender_select_module) that was added in Xliff 2.2 rather than the `<unit mf:select>` and `<segment id>` attributes.

The `pgs:case` values use a `*:` prefix for the default case, as in `*:other` or `*:nominative`. This is a deliberate difference from the Xliff spec's [recommendation](https://docs.oasis-open.org/xliff/xliff-core/v2.2/xliff-extended-v2.2-part2.html#pgs_case:~:text=should%20also%20match), to allow for the identifier of the default case to be retained.

CC @srl295, @catamorphism

### DIFF
--- a/mf2/xliff/src/mf2-to-xliff.ts
+++ b/mf2/xliff/src/mf2-to-xliff.ts
@@ -14,8 +14,6 @@ import { toNmtoken } from './nmtoken.ts';
 let _id = 0;
 const nextId = () => String(++_id);
 
-const star = Symbol('*');
-
 // TODO: Support declarations
 export function mf2xliff(
   source: MessageFormatInfo,
@@ -51,16 +49,10 @@ export function mf2xliff(
   return { type: 'element', name: 'xliff', attributes, elements: [file] };
 }
 
-function msgId(
-  pre: 'f' | 'g' | 's' | 'u',
-  key: string[],
-  variant?: (string | typeof star)[]
-) {
+function msgId(pre: 'f' | 'g' | 's' | 'u', key: string[], variant?: string[]) {
   const id = `${pre}:${key.map(toNmtoken).join('.')}`;
   return variant?.length
-    ? `${id}:${variant
-        .map(v => (typeof v === 'string' ? toNmtoken(v) : '_other'))
-        .join('.')}`
+    ? `${id}:${variant.map(v => toNmtoken(v.replace('*:', ''))).join('.')}`
     : id;
 }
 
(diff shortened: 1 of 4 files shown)


srl295:
> Fixes #456
> 
> This adopts the attributes introduced in the [Plural, Gender, and Select Module](https://docs.oasis-open.org/xliff/xliff-core/v2.2/xliff-extended-v2.2-part2.html#plural_gender_select_module) that was added in Xliff 2.2 rather than the `<unit mf:select>` and `<segment id>` attributes.

Great!

> The `pgs:case` values use a `*:` prefix for the default case, as in `*:other` or `*:nominative`. This is a deliberate difference from the Xliff spec's [recommendation](https://docs.oasis-open.org/xliff/xliff-core/v2.2/xliff-extended-v2.2-part2.html#pgs_case:~:text=should%20also%20match), to allow for the identifier of the default case to be retained.

I'm trying to understand this one.  Why does the "identifier of the default case" need to be retained, at least for plurals, when "other" is always the default case? Or are there situations where `other` would not be the default plural?

Maybe answering my own question: Are you trying to preserve the distinction between `*` and `other`? I think that's not an important distinction for plurals, in fact, I'd think that `*` (and not `other`) should be used in mf2.  

What about preserving the non-compliant `*:other` only if someone passes in `other`, and using just `other` if `*` is used?

srl295 (approved):
Looks good though a question about `*:other`

eemeli:
> > The `pgs:case` values use a `*:` prefix for the default case, as in `*:other` or `*:nominative`. This is a deliberate difference from the Xliff spec's [recommendation](https://docs.oasis-open.org/xliff/xliff-core/v2.2/xliff-extended-v2.2-part2.html#pgs_case:~:text=should%20also%20match), to allow for the identifier of the default case to be retained.
> 
> I'm trying to understand this one. Why does the "identifier of the default case" need to be retained, at least for plurals, when "other" is always the default case? Or are there situations where `other` would not be the default plural?

@srl295 While the MF2 syntax does not support default case labels, the [data model does](https://github.com/unicode-org/message-format-wg/tree/main/spec/data-model#:~:text=value%3F%3A%20string%3B). It's mostly there to support MF2 being used to represent other syntaxes that do support such labels, like Fluent. The feature itself has a couple of use cases:
1. Selection on grammatical case, or similar situations where it's useful for a translator to know the actual title for each case, e.g. `*:nominative` rather than `other`.
2. Selection on enums or other sets which do not contain a natural "other". A minimal example of this is selection on a default-true boolean value, where `false` and `*:true` are much more useful than `false` and `other`.

> Maybe answering my own question: Are you trying to preserve the distinction between `*` and `other`? I think that's not an important distinction for plurals, in fact, I'd think that `*` (and not `other`) should be used in mf2.

For plurals in particular, it's also good to be able to avoid `other` as a label, in particular for Slavic languages. For example, `:integer` selection in [Polish](https://www.unicode.org/cldr/charts/48/supplemental/language_plural_rules.html#pl) will only ever resolve to `one`, `few`, or `many`, as `other` is only used for fractional values. In MF2 syntax we avoid this by the cases for that being `one`, `few`, and `*`.

For Xliff I think we do want to present some textual label here, as we should not presume that either the translator or the tooling working with the file will have any understanding or familiarity with MF2 or its syntax.

> What about preserving the non-compliant `*:other` only if someone passes in `other`, and using just `other` if `*` is used?

Yeah, defaulting to just `other` does make sense when there isn't a literal "other" case and when no explicit label is provided for the default case. That should probably come as a separate change, though, as it might mean even more finangling of the selection generation. It's been a while since I wrote it, and it gets a bit complicated to account for cases where the source and target use different selectors.

